Issue #15, opened 2019-10-23
Set up GitHub Actions workflow for build and publish pkg


youngbupark:
## Describe the proposal

Set up GitHub Actions workflow for build and publish pkg

* [ ] Create Workflow for build
* [ ] Create workflow for publishing pkg

ksivamuthu:
@youngbupark Can you please assign this to me? I will work on it. Thanks

PR #27, opened 2020-03-28
Adding npm build/Publish GitHub actions

ksivamuthu:
# Description

1. Add npm build workflow
2. Add npm publish on release creation

Todo:
* [ ] Add NPM_TOKEN to secrets
* [ ] Publish to Github Package Registry

Now the release process is, 

1. Update package.json with the version to publish
2. Create the release with the same version
3. Once the release is created, the npm package will be published to the npm registry.

It can be enhanced n various ways

1. Creating release will automatically bump version in package.json and commit/push from github actions.
2. The second option is creating an automatic release and publish, once the tag is created.

Let me know which will work better for dapr or any other workflows.

## Issue reference

This PR will close: #15 

## Checklist

* [x] Code compiles correctly
* [ ] Created/updated tests


.github/workflows/npmpublish.yml
```diff
@@ -0,0 +1,32 @@
+# This workflow will run tests using node and then publish a package to GitHub Packages when a release is created
```
pruthvidhodda, line 1: nit: this publishes packages to npm registry

.github/workflows/nodejs.yml
```diff
@@ -0,0 +1,31 @@
...
+        node-version: [10.x, 12.x]
+
+    steps:
+    - uses: actions/checkout@v2
+    - name: Use Node.js ${{ matrix.node-version }}
+      uses: actions/setup-node@v1
+      with:
+        node-version: ${{ matrix.node-version }}
+    - run: npm ci
+    - run: npm run build --if-present
```
pruthvidhodda, line 29: Why do we need this?

pruthvidhodda:
@youngbupark @amanbha  Should we start using [npm-scope](https://docs.npmjs.com/using-npm/scope.html) for our packages?

.github/workflows/npmpublish.yml
```diff
@@ -0,0 +1,32 @@
...
+# For more information see: https://help.github.com/actions/language-and-framework-guides/publishing-nodejs-packages
+
+name: Node.js Package
+
+on:
+  release:
+    types: [created]
+
+jobs:
+  build:
```
pruthvidhodda, line 11: Do we need this job when we are anyways running `npm ci` in publish-npm job?

ksivamuthu: It came from the default template of github actions to do the test before publishing it. Since this dapr node lib doesn't have build/test commands, we can remove it. I will do.

.github/workflows/npmpublish.yml
ksivamuthu: Same - Comments from the default template. Do we've any plans on releasing in the github packages too?

.github/workflows/nodejs.yml
ksivamuthu: I'll remove. We don' have build/test command in this library

ksivamuthu:
General comments: Does dapr have plan to do lint and add a unit tests in the npm module? For e.g  #28 to capture format/js issues like this that causes an error

.github/workflows/npmpublish.yml
pruthvidhodda: I think it's better if we publish to github packages as well, that would add one more job to the workflow, do we have any other concerns? Are you planning to do it as part of the same PR? If not we can raise an issue here: https://github.com/dapr/js-sdk/issues

.github/workflows/nodejs.yml
```diff
@@ -0,0 +1,30 @@
+# This workflow will do a clean install of node dependencies, build the source code and run tests across different versions of node
+# For more information see: https://help.github.com/actions/language-and-framework-guides/using-nodejs-with-github-actions
+
+name: Node.js CI
```
youngbupark, line 4: I would merge nodejs and npmpublish to single build.yaml. publish needs to be conditional.

please refer to this work flow - https://github.com/dapr/java-sdk/blob/master/.github/workflows/build.yml

publish will be done only if maintainer tags v-[version]

amanbha: yes, publish would happen conditionally or it might be manual initially.

amanbha:
> General comments: Does dapr have plan to do lint and add a unit tests in the npm module? For e.g #28 to capture format/js issues like this that causes an error

Yes, it should be enabled except for the grpc generated code.

ksivamuthu:
I ll work on it today. Sorry for the delay.

amanbha:
> I ll work on it today. Sorry for the delay.

no worries, thanks for the help!

ksivamuthu:
@amanbha @youngbupark Added lint and updated conditional publish in github actions. Need to add Github/NPM token in secrets. Thanks for your patience.

youngbupark:
> @amanbha @youngbupark Added lint and updated conditional publish in github actions. Need to add Github/NPM token in secrets. Thanks for your patience.

Thanks @ksivamuthu for your contribution! I have small comment on it. can we rename nodejs.yaml file name to build.yaml ? when we add build badge, GH actions shows the workflow file name... so I would like to see `build` inside badge rather than `nodejs` :)

ksivamuthu:
> > @amanbha @youngbupark Added lint and updated conditional publish in github actions. Need to add Github/NPM token in secrets. Thanks for your patience.
> 
> Thanks @ksivamuthu for your contribution! I have small comment on it. can we rename nodejs.yaml file name to build.yaml ? when we add build badge, GH actions shows the workflow file name... so I would like to see `build` inside badge rather than `nodejs` :)

Renamed.

youngbupark:
@ksivamuthu let me add the secrets once this pr is merged.

Issue #32, opened 2020-05-26
Update protobuf.sh to work with changes to proto files in dapr/dapr


pruthvidhodda:
## Describe the proposal
The proto files in dapr/dapr are renamed and files are moved to new folder (runtime). Update protobuf.sh file to work with these changes.

PR #33, opened 2020-05-26
Fix protobuf.sh

pruthvidhodda:
# Description

Updated protobuf.sh to work with changes to proto files in dapr/dapr

## Issue reference

Please reference the issue this PR will close: #32 

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests


protobuf.sh
```diff
@@ -41,11 +42,12 @@ checkHttpRequestCLI() {
 
 downloadFile() {
 
-    FILE_NAME=$1
-    FILE_PATH="${PROTO_PATH}/${FILE_NAME}/v1"
+    FOLDER_NAME=$1
+    FILE_NAME=$2
+    FILE_PATH="${PROTO_PATH}/${FOLDER_NAME}/v1"
```
vinayada1, line 47: Doesn't this need to be: FILE_PATH="${PROTO_PATH}/${FOLDER_NAME}/v1/{FILE_NAME}" ?

pruthvidhodda: We are using FILE_PATH to create directories as needed in that path. Having filename there would create another folder with name {FILE_NAME} which we do not want to happen.

PR #77, opened 2021-03-26
Update proto clients in js-sdk

pkedy:
# Description

* Version bump to `0.12.0-preview01`
* Ran `./protobuf.sh` to update protos to align with the Dapr core 1.1 release
* Added components for the sample to work and updated README

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "dapr-client",
-  "version": "0.12.0",
+  "version": "0.12.0-preview01",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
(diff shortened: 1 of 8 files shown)


package-lock.json
```diff
@@ -1,8 +1,4540 @@
 {
   "name": "dapr-client",
-  "version": "0.9.0-preview01",
-  "lockfileVersion": 1,
+  "version": "0.12.0",
```
artursouza, line 3: It is still preview at this point. We remove the RC on the day of release. On the other hand, given the scope of this SDK, it might be OK to skip and release the minor version right away.

artursouza (approved):

tcnghia (approved):

Issue #136, opened 2021-11-16
Actors: Should not allow state manipulation outside Actor Runtime


XavierGeerinck:
## Expected Behavior

We should not be able to modify the state of actors through the DaprClient

PR #137, opened 2021-11-17
feat(actors): implement actor proxy object and remove actor access remotely

XavierGeerinck:
# Description

This PR implements the Actor Proxy object and removes the possibility to access actors and its methods remotely.

## Issue reference

Please reference the issue this PR will close: #136 

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests
* [x] Extended the documentation


greenie-msft:
Thank you, Xavier for the refactor. @artursouza and @halspang please review when you have a moment.

src/actors/client/ActorClient/ActorClient.ts
```diff
@@ -0,0 +1,36 @@
...
+import IClient from '../../../interfaces/Client/IClient';
+import IClientActor from '../../../interfaces/Client/IClientActor';
+import ActorClientGRPC from './ActorClientGRPC';
+import ActorClientHTTP from './ActorClientHTTP';
+
+export default class ActorClient {
+  readonly daprClient: DaprClient;
+  readonly actor: IClientActor;
+
+  constructor(daprClient: DaprClient) {
```
artursouza, line 14: What is the advantage of passing this object here just to get it later? Is there a way to get the host and port directly and avoid requiring the actual general purpose Dapr client?

src/actors/client/ActorClient/ActorClientGRPC.ts
```diff
@@ -0,0 +1,278 @@
...
+
+// https://docs.dapr.io/reference/api/actors_api/
+export default class ActorClientGRPC implements IClientActor {
+  client: GRPCClient;
+
+  constructor(client: GRPCClient) {
+    this.client = client;
+  }
+
+  async invoke(method: "GET" | "POST" | "PUT" | "DELETE", actorType: string, actorId: string, methodName: string, body?: any): Promise<object> {
```
artursouza, line 19: method seems to be unused, which makes sense because it is always Post.

src/actors/client/ActorClient/ActorClientGRPC.ts
```diff
@@ -0,0 +1,278 @@
...
+      client.invokeActor(msgService, (err, res: InvokeActorResponse) => {
+        if (err) {
+          return reject(err);
+        }
+
+        // https://docs.dapr.io/reference/api/secrets_api/#response-body
+        const resData = Buffer.from(res.getData()).toString();
+
+        try {
+          return resolve(JSON.parse(resData));
```
artursouza, line 41: We can start with JSON but I highly recommend making the serialization a pluggable interface.

src/actors/client/ActorClient/ActorClientGRPC.ts
```diff
@@ -0,0 +1,278 @@
...
+        }
+
+        // https://docs.dapr.io/reference/api/actors_api/#http-response-codes-3
+        return resolve();
+      });
+    });
+  }
+
+  // @todo: not implemented, cannot find the gRPC bindings
+  // async reminderGet(actorType: string, actorId: string, name: string): Promise<void> {
```
artursouza, line 140: There is no reason to implement this. Please, remove the code. Once registered, reminders are kept in the runtime.

src/actors/client/ActorClient/ActorClientGRPC.ts
```diff
@@ -0,0 +1,278 @@
...
+          return reject(err);
+        }
+
+        // https://docs.dapr.io/reference/api/actors_api/#http-response-codes-3
+        return resolve();
+      });
+    });
+  }
+
+  async timerDelete(actorType: string, actorId: string, name: string): Promise<void> {
```
artursouza, line 225: I wonder if these methods should be called deleteTimer, createTimer, etc. We went through an exercise to normalize naming among the SDKs a while ago and only accept variations that are idiomatic to each language. See the example in Java SDK:

https://github.com/dapr/java-sdk/blob/master/sdk-actors/src/main/java/io/dapr/actors/runtime/AbstractActor.java

src/actors/client/ActorClient/ActorClientGRPC.ts
```diff
@@ -0,0 +1,278 @@
...
+        try {
+          return resolve(JSON.parse(resData));
+        } catch (e) {
+          return resolve(resData as any);
+        }
+      });
+    });
+  }
+
+  async stateTransaction(actorType: string, actorId: string, operations: OperationType[]): Promise<void> {
```
artursouza, line 49: In the implementation in the Java SDK, we keep a separate client for the "server" side of actor and the "client" side of actors. It basically means that the invoke API is in a separate interface and implementation. It is important to keep the abstraction between client and server.

src/actors/client/ActorProxy.ts
```diff
@@ -1,38 +0,0 @@
-import { AbstractActor } from "../..";
-import DaprClient from "../../implementation/Client/DaprClient";
-import Class from "../../types/Class";
-import ActorId from "../ActorId";
-import ActorProxyFactory from "./ActorProxyFactory";
-
-export default class ActorProxy<T extends AbstractActor> {
```
artursouza, line 7: Why are you removing ActorProxy? Also, this would be a breaking change, not?

src/actors/client/ActorProxyFactory.interface.ts
```diff
@@ -1,8 +0,0 @@
-import Class from "../../types/Class";
-import ActorId from "../ActorId";
-import ActorProxy from "./ActorProxy";
-import AbstractActor from "../runtime/AbstractActor";
-
-export default interface ActorProxyFactoryInterface {
```
artursouza, line 6: breaking change?

src/actors/client/ActorProxyFactory.ts
```diff
@@ -1,17 +0,0 @@
-import { AbstractActor, DaprClient } from "../..";
-import Class from "../../types/Class";
-import ActorId from "../ActorId";
-import ActorProxy from "./ActorProxy";
-import ActorProxyFactoryInterface from "./ActorProxyFactory.interface";
-
-export default class ActorProxyFactory implements ActorProxyFactoryInterface {
```
artursouza, line 7: breaking change?

src/implementation/Client/GRPCClient/state.ts
```diff
@@ -180,4 +182,33 @@ export default class GRPCClientState implements IClientState {
       });
     });
   }
+
+  async query(storeName: string, query: StateQueryType): Promise<StateQueryResponseType> {
```
artursouza, line 186: Please, do not bundle the query API changes in this PR. This can be merged independently. BTW, there is a breaking change in this API already in 1.6, please sync with @dmitsh

src/implementation/Client/HTTPClient/HTTPClient.ts
```diff
@@ -43,6 +43,11 @@ export default class HTTPClient implements IClient {
     return CommunicationProtocolEnum.HTTP;
   }
 
+  async executeWithApiVersion(apiVersion: string = "v1.0", url: string, params: any = {}): Promise<object | string> {
```
artursouza, line 46: separate PR?

artursouza (changes requested):
Please, consider separating the query state API changes in a new PR.

src/actors/client/ActorClient/ActorClient.ts
XavierGeerinck: Well, the DaprClient could be transformed into a Singleton if that makes more sense? In the end there should be only one active throughout the code. Would that make sense?

src/actors/client/ActorClient/ActorClientGRPC.ts
XavierGeerinck: I do agree that it might be easier to have default POST and move method to the end as optional

src/actors/client/ActorClient/ActorClientGRPC.ts
XavierGeerinck: Let's skip for this PR?

src/actors/client/ActorClient/ActorClientGRPC.ts
XavierGeerinck: Ok will change

src/actors/client/ActorProxy.ts
XavierGeerinck: There is a ProxyBuilder now that implements Proxy itself. This was old deprecated code and was not functional

src/actors/client/ActorProxyFactory.interface.ts
XavierGeerinck: There is a ProxyBuilder now that implements Proxy itself. This was old deprecated code and was not functional

src/actors/client/ActorProxyFactory.ts
XavierGeerinck: There is a ProxyBuilder now that implements Proxy itself. This was old deprecated code and was not functional

src/implementation/Client/GRPCClient/state.ts
XavierGeerinck: Personally, rather than extracting this I would prefer to merge it now and fix it afterwards in a new PR. Bit difficult to get it out now :) thoughts?

src/implementation/Client/HTTPClient/HTTPClient.ts
XavierGeerinck: Personally, rather than extracting this I would prefer to merge it now and fix it afterwards in a new PR. Bit difficult to get it out now :) thoughts?

src/actors/client/ActorClient/ActorClient.ts
artursouza: Just let the implementation of the ActorClient to solve how to keep a reference to DaprClient. Not every actor implementation should need this instance.

src/actors/client/ActorClient/ActorClientGRPC.ts
artursouza: Please, keep POST or PUT only. No need to expose this. Here is why:

Java does POST only: https://github.com/dapr/java-sdk/blob/master/sdk-actors/src/main/java/io/dapr/actors/client/DaprHttpClient.java
.Net is using PUT only: https://github.com/dapr/dotnet-sdk/blob/0c9d6a45c8d3792a92d7141056c390bea098d02b/src/Dapr.Actors/DaprHttpInteractor.cs#L199

Runtime does indeed accept all these HTTP Methods: https://github.com/dapr/dapr/blob/ab1c6a0ac7041ff9da34626e05b3628277f61acb/pkg/http/api.go#L297
On the other hand, when invoking actor locally, we only use PUT. So, the method you use in the SDK does not matter in the end. This also allows the gRPC API to work seamlessly.

src/actors/client/ActorClient/ActorClientGRPC.ts
artursouza: Yes. This is a known issue in the SDKs but "JSON" is mostly compatible cross languages - unless the object has Date/Time types or something that is not portable.

src/implementation/Client/GRPCClient/state.ts
artursouza: Well. If the code is in separate files, separating is not that hard. Also, if you have it in separate commits, you can remove the commit via a rebase and then cherry-pick it in a new branch.

src/implementation/Client/HTTPClient/HTTPClient.ts
artursouza: I think it is doable.

artursouza (changes requested):

src/actors/client/ActorClient/ActorClientGRPC.ts
XavierGeerinck: Ok, I will remove the method

src/implementation/Client/GRPCClient/state.ts
XavierGeerinck: Removed it

src/implementation/Client/HTTPClient/HTTPClient.ts
XavierGeerinck: Removed it

src/actors/client/ActorClient/ActorClient.ts
XavierGeerinck: Ah now I see it, apologies, was sleeping I think when writing this 😅😂

greenie-msft:
@XavierGeerinck @artursouza - any update on this PR? Would be great to get this in soon.

XavierGeerinck:
Changes were submitted awaiting final review normally :)

src/actors/runtime/AbstractActor.ts
```diff
@@ -200,7 +204,11 @@ export default abstract class AbstractActor {
     return this.stateManager;
   }
 
-  getId(): ActorId {
+  getId(): string {
```
artursouza, line 207: Can we keep the old way? It feels redundant to have these two variants.

artursouza (changes requested):
Just a small change.

Issue #140, opened 2021-11-23
Feature: Configuration API


XavierGeerinck:
## Describe the proposal

https://docs.dapr.io/developing-applications/building-blocks/configuration/

Will also close #258

XavierGeerinck:
Assigned tanvigour  as requested in #143

XavierGeerinck:
https://github.com/dapr/dapr/issues/2988#issuecomment-881511141

tanvigour:
Hey @XavierGeerinck Is this being worked on currently?

XavierGeerinck:
Hi @tanvigour this is, but we are awaiting a dapr/dapr release to resolve issues in tests

PR #154, opened 2022-01-17
Initial Tests and Code for Configuration API

XavierGeerinck:
Signed-off-by: Xavier Geerinck <xavier.geerinck@gmail.com>

# Description

Implemented Configuration API

## Issue reference

Please reference the issue this PR will close: #140 

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests

## Open Questions

- [ ] AppId, why is this used? https://github.com/dapr/dapr/issues/2988#issuecomment-881511141 shows GetConfigurationRequest with appId but is not used in implementation at https://github.com/dapr/dapr/blob/master/dapr/proto/runtime/v1/dapr.proto#L90
- [ ] When I set metadata, it doesn't seem to be passed? (I am not receiving it back even though the code shows that I am setting it?) Java SDK doesn't seem to check for it either? https://github.com/dapr/java-sdk/commit/185cdba293de21f92855815792ce936c5164691a#diff-eab2bac5cf4c2f6deea0453b132eada833a4781dc96dcc09aa625d32464978c9R105
- [ ] When calling subscribe, I expect to listen to all key changes. However when having an empty key list, it doesn't listen to anything? Very odd
- [ ] Stream gets opened, but what if we open multiple streams?


artursouza:
Let me try to answer your questions:

1. We don't use appId. That was discussed in some thread.
2. Metadata is to modify behavior in the component, not necessarily to be returned back. Returned metadata is only if component implements it.
3. I agree. Please, cut an issue.
4. Can you clarify the context for the last question?

XavierGeerinck:
> Let me try to answer your questions:
> 
> 1. We don't use appId. That was discussed in some thread.
> 2. Metadata is to modify behavior in the component, not necessarily to be returned back. Returned metadata is only if component implements it.
> 3. I agree. Please, cut an issue.
> 4. Can you clarify the context for the last question?

1. Ok
2. Ok
3. Will do
4. When I open 1 stream, I am unable to open a 2nd one. It doesn't start subscribing to updates anymore then (second test fails, first passes)

XavierGeerinck:
Note: awaiting https://github.com/dapr/dapr/pull/4189

Issue #178, opened 2022-02-07
JS-SDK client uses environment variables to initialize


paulyuk:
## Describe the proposal
Currently the Dapr sidecar is setting environment variables on your behalf like DAPR_HOST and DAPR_PORT.  The proposal is to have a default constructor for the DaprClient() that sets the internal host, port, and API version using these environment variables.  

This has the benefits of:
- consistency with other SDKs like Python and .NET: [example](https://github.com/dapr/python-sdk/blob/master/dapr/clients/http/client.py)
- Less user code to get the best practice

Upon completion the user should be able to successfully write this code with parameterless ctor to successfully connect to sidecar and call APIs:
```
const client = new DaprClient();
await client.pubsub.publish("order_pub_sub", "orders", {orderId: 100});
```

XavierGeerinck:
This should be possible as https://docs.dapr.io/reference/environment/ expoes `DAPR_HTTP_PORT`, `DAPR_GRPC_PORT` and `APP_PORT`. Would need to check if when running `dapr run` this is set as well.

shubham1172:
I see `DAPR_API_VERSION` in python-sdk [here](https://github.com/dapr/python-sdk/blob/e167ec089cb8a614f675ee799b4aa978d1422852/dapr/conf/global_settings.py#L25), but this is not present in any other SDK or CLI/Runtime repositories. It's not documented here too - https://docs.dapr.io/reference/environment/

This does not look like a supported feature. @paulyuk, thoughts?

Also, since we are refactoring `DaprClient`, we can refactor `DaprServer` also under the same work-item.

PR #192, opened 2022-02-16
Use environment variables to initialize client and server

shubham1172:
# Description

This PR makes `DaprClient` and `DaprServer` configurable using environment variables. If missing, they fallback to a sane default value. 

## Issue reference

Please reference the issue this PR will close: #178 

## Checklist

* [X] Code compiles correctly
* [x] Created/updated tests

### DIFF
--- a/src/implementation/Client/DaprClient.ts
+++ b/src/implementation/Client/DaprClient.ts
@@ -53,14 +53,14 @@ export default class DaprClient {
   readonly actor: IClientActorBuilder;
 
   constructor(
-    daprHost = Settings.getDefaultHost()
+    daprHost?: string
     , daprPort?: string
     , communicationProtocol: CommunicationProtocolEnum = CommunicationProtocolEnum.HTTP
     , options: DaprClientOptions = {
       isKeepAlive: true
     }
   ) {
-    this.daprHost = daprHost;
+    this.daprHost = daprHost ?? Settings.getDefaultHost();
     this.daprPort = daprPort ?? Settings.getDefaultPort(communicationProtocol);
     this.communicationProtocol = communicationProtocol;
     this.options = options;
--- a/src/implementation/Client/DaprClient.ts
+++ b/src/implementation/Client/DaprClient.ts
@@ -53,14 +53,14 @@ export default class DaprClient {
   readonly actor: IClientActorBuilder;
 
   constructor(
-    daprHost?: string
+    daprHost = Settings.getDefaultHost()
     , daprPort?: string
     , communicationProtocol: CommunicationProtocolEnum = CommunicationProtocolEnum.HTTP
     , options: DaprClientOptions = {
       isKeepAlive: true
     }
   ) {
-    this.daprHost = daprHost ?? Settings.getDefaultHost();
+    this.daprHost = daprHost;
     this.daprPort = daprPort ?? Settings.getDefaultPort(communicationProtocol);
     this.communicationProtocol = communicationProtocol;
     this.options = options;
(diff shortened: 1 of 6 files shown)


src/implementation/Client/HTTPClient/HTTPClient.ts
```diff
@@ -16,12 +16,15 @@ export default class HTTPClient implements IClient {
...
+  constructor(
+    host = Settings.getDefaultHost()
+    , port = Settings.getDefaultHttpPort()
+    , options: DaprClientOptions = {
+      isKeepAlive: true
+    }
+  ) {
     this.clientHost = host;
-    this.clientPort = port;
+    this.clientPort = port ?? port;
```
XavierGeerinck, line 27: ? What does this do? port or port?

src/implementation/Client/DaprClient.ts
```diff
@@ -52,15 +53,15 @@ export default class DaprClient {
   readonly actor: IClientActorBuilder;
 
   constructor(
-    daprHost: string
-    , daprPort: string
+    daprHost = Settings.getDefaultHost()
```
XavierGeerinck, line 56: I am not a fan personally of this line, the below follows a different way of working than this. Could we make it so constructor has all optional parameters then and we add the same `??` to daprHost as welll on line 63?

src/implementation/Client/HTTPClient/HTTPClient.ts
shubham1172: Oops, thanks for catching that. Only supposed to be `port`.

src/implementation/Client/DaprClient.ts
shubham1172: Good suggestion, that will look more uniform!

shubham1172: For `GRPCClient` and `HTTPClient`, I have still kept it inline as that looks easier to read. Let me know what your thoughts are, we can change that too.

src/implementation/Server/DaprServer.ts
```diff
@@ -35,19 +36,19 @@ export default class DaprServer {
...
     , clientOptions: DaprClientOptions = {
       isKeepAlive: true
     }
   ) {
-    this.serverHost = serverHost;
-    this.serverPort = serverPort;
-    this.daprHost = daprHost;
-    this.daprPort = daprPort;
+    this.serverHost = serverHost ?? Settings.getDefaultHost();
+    this.serverPort = serverPort ?? Settings.getDefaultAppPort(communicationProtocol);
```
shubham1172, line 49: Also here, should we check for `DAPR_SERVER_PORT`? I don't see it documented anywhere, but we do set that environment variable only in `DaprServer`.

src/implementation/Client/DaprClient.ts
XavierGeerinck: Ok for me

src/implementation/Server/DaprServer.ts
XavierGeerinck: No I wouldn't set it. Let's stick to the Dapr documented one, let's work towards deprecating the DAPR_SERVER_PORT in the future as well?

shubham1172: Cool, we don't have `DAPR_SERVER_PORT` in the new code. Resolving this.

shubham1172:
@XavierGeerinck, I had seen `??` in other TS codebases. Also saw this one https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#nullish-coalescing 
image

From https://stackoverflow.com/questions/61480993/when-should-i-use-nullish-coalescing-vs-logical-or, `||` 
> But the OR operator || can be problematic if your left value might contain "" or 0 or false

What do you think?

XavierGeerinck:
Awesome, LGTM

PR #200, opened 2022-02-23
[WIP] Update testing steps in CI

### LINKED ISSUE #193: Code coverage reporting
This proposal is to introduce code coverage reporting for PRs.

### LINKED ISSUE #197: Pipeline passes even though e2e test fails
## Expected Behavior

If an e2e test fails, the pipeline should also fail.

## Actual Behavior
Test failure:
image
Pipeline passing:
image

Reference: https://github.com/dapr/js-sdk/runs/5299603320?check_suite_focus=true

## Steps to Reproduce the Problem
Fail any e2e test

shubham1172:
# Description

Sorry for slightly multi-functional PR. The changes are closely related to each other. It does the following:

1. Separates unit tests from e2e tests - unit tests are now a part of the build pipeline. 
2. Fixes e2e testing issue (#197)
3. Introduces codecov (#193)
4. Adds `utils` unit tests to `package.json`

## Issue reference

Please reference the issue this PR will close #193 and close #197

## Checklist

* (NA) Code compiles correctly
* (NA) Created/updated tests
* (NA) Extended the documentation


shubham1172:
@XavierGeerinck, it's currently a WIP, I was trying out different things. 

The first approach is something similar to what conformance tests in https://github.com/dapr/components-contrib do. We can set an environment variable on failure, and use that to detect in the next step. This is a hack, but should be a good short term solution till we reach the second approach. Also, the code is commented because the CI started going crazy in an infinite loop, didn't want it to break, LOL.

The second approach will be like what Java SDK/.NET SDK do today. Reference: https://github.com/dapr/java-sdk/tree/master/sdk-tests/src/test/java/io/dapr/it. They have an elaborate test runner, that is responsible for starting up Dapr 
 on random ports, with timeouts and retries in place. It listens for DAPR logs, and looks for "success criteria". The test runner invokes `dapr run` with command to run the tests (similar to our `test:e2e:all` in package.json), but since the control is with the test runner now, we are free to check the logs and decide if the tests passed. 

The second approach is better than first, but will take more time. I think once this is fixed, https://github.com/dapr/cli/issues/684, our original method should work, but that change might require a lot of time since people would already have dependencies in place.

shubham1172:
As discusses offline, we are leaning to provide a short-term fix using first approach, and then plan the second approach as a long-term fix.

Issue #208, opened 2022-03-07
Improve Dapr-Bot


XavierGeerinck:
## Describe the proposal

We could improve the Dapr Bot for Repo Management with the following:

* `stale is fine`: Mark an issue as stale when it is not being responded to in X days
* `needs-author-feedback`: Add the tag when a comment is made with `/ping-author` by a contributor. This will then add the label `needs-author-feedback` to the issue.
* `needs-team-attention`: Add the tag when the Author of the issue replies to a comment that added the label `needs-author-feedback`. It will as well remove the tag `needs-author-feedback`

shubham1172:
Additionally, add a `/ok-to-e2e-test` command.

tanvigour:
> Additionally, add a `/ok-to-e2e-test` command.

Doesn't ok-to-test do that?

shubham1172:
@tanvigour, we don't have these currently in this repo's dapr bot.

PR #210, opened 2022-03-10
Enhance Dapr Bot

shubham1172:

# Description

Adds a few more features to the dapr bot
1. Cron job to mark issues and PR stale
2. `/ok-to-e2e-test` to retrigger end to end tests
3. `/ping-author` to add special label `needs-author-feedback`
4. Replacing `needs-author-feedback` with `needs-team-attention` when author replies after (3)

## Issue reference

Please reference the issue this PR will close: #208 

## Checklist

* (NA) Code compiles correctly
* (NA) Created/updated tests
* (NA) Extended the documentation


tanvigour (commented):
I think we should also have something for better stale management of issues. For example, if there's no activity for say 3 days, ping a maintainer?

shubham1172:
@tanvigour I can think of two scenarios here.
1. User opens an issue saying "Help with XYZ" and when we ask for more context, they ghost us. In this case, the action item is on them, and stale bot will simply mark stale/close it after threshold days.
2. There is a feature request/bug which requires action item from a contributor/maintainer. In this case, the stale bot should not close the issue. This can be prevented by adding special labels to the issue (mentioned in the workflow).

There is another aspect to it. Maintainers and contributors are responsible for regularly triaging the issues. They also have their GitHub notifications already hot, so any ping we might add will add to the noise. Plus, it is possible that the action item is already on a single maintainer, so in that case, how do we ping only the person who's responsible? IMO, we can skip this feature for now and manually keep an eye on issues. Case (1) above is the purpose of the stale bot, and Case (2) can be avoided  by regular triaging of the repository (the maintainer can simply add the label `triaged/resolved`).

Also, this repository does not see a huge flux of issues. The current count is also very less. Chances are that each issue is already under a contributor's radar.

Please feel free to share your thoughts - @XavierGeerinck @amulyavarote

XavierGeerinck:
I agree with you @shubham1172 I would however tag wise look at the `needs-author-feedback` and `needs-team-attention` labels instead of `triaged/resolved`? Seeing that those are the "`ping/pong`" mechanism and only ping us whenever we know there is an action at our side. It will also prevent the stale bot removing those issues

tanvigour (approved):

amulyavarote (dismissed):

shubham1172:
@XavierGeerinck 

> I agree with you @shubham1172 I would however tag wise look at the `needs-author-feedback` and `needs-team-attention` labels instead of `triaged/resolved`? Seeing that those are the "`ping/pong`" mechanism and only ping us whenever we know there is an action at our side. It will also prevent the stale bot removing those issues

When an issue is tagged with `needs-author-feedback`, we should still be able to mark them stale, right? Say someone posted an issue where they are facing XYZ issues and we ask for more information, adding the tag `needs-author-feedback`, but they don't get back in 60 days.

.github/scripts/dapr_bot.js
```diff
@@ -0,0 +1,149 @@
+/**
+ * Execute fn if label exists on an issue.
+ * @param {*} github GitHub object reference
+ * @param {*} label label name
+ * @param {*} issue GitHub issue reference
+ * @param {*} fn async function
+ */
+ async function executeIfIssueHasLabel(github, issue, label, fn) {
+    var response = await github.issues.listLabelsOnIssue({
```
XavierGeerinck, line 9: I would use `const`

.github/scripts/dapr_bot.js
```diff
@@ -0,0 +1,149 @@
...
+        issue_number: issue.number,
+        owner: issue.owner,
+        repo: issue.repo,
+    });
+
+    var labelNames = response.data.map((i) => i.name)
```
XavierGeerinck, line 15: `const`

.github/scripts/dapr_bot.js
```diff
@@ -0,0 +1,149 @@
...
+                owner: issue.owner,
+                repo: issue.repo,
+                labels: ['needs-author-feedback']
+            })
+            return;
+        }
+    }
+
+    // commands for GH pull requests
+    if (isFromPulls && commentBody) {
```
XavierGeerinck, line 120: 1 if statement?

XavierGeerinck (changes requested):
Overal "fine" some minor style changes.

Big question I have: Wouldn't it be easier and more readable if we process by comment body and respective functions?

E.g. (pseudo wise)

```
switch (commentBody) {
    case "/assign":
        executeAssign();
        break;
    case "/other":
        executeOther();
        break;
}
```

.github/scripts/dapr_bot.js
```diff
@@ -0,0 +1,149 @@
...
+    var response = await github.issues.listLabelsOnIssue({
+        issue_number: issue.number,
+        owner: issue.owner,
+        repo: issue.repo,
+    });
+
+    var labelNames = response.data.map((i) => i.name)
+    for (const labelName of labelNames) {
+        if (labelName == label) {
+            await fn()
```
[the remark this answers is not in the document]
shubham1172: This is better! Thanks!

shubham1172:
> Overal "fine" some minor style changes.
> 
> Big question I have: Wouldn't it be easier and more readable if we process by comment body and respective functions?
> 
> E.g. (pseudo wise)
> 
> ```
> switch (commentBody) {
>     case "/assign":
>         executeAssign();
>         break;
>     case "/other":
>         executeOther();
>         break;
> }
> ```

It was mostly a port of dapr/dapr earlier. I have refactored it - but it comes with some boilerplates. This is more readable for sure but. Let me know if this works!

XavierGeerinck (approved):

Issue #201, opened 2022-02-25
Fixing dependencies with bad license


shubham1172:
Main issue: https://github.com/dapr/dapr/issues/4236

Please refer to the parent issue to help with fixing existing issues with bad licenses. 

This is required for Dapr to be compliant with CNCF guidelines, so please add this issue to the SDK planning accordingly.

shubham1172:
@amulyavarote please refer to https://github.com/dapr/dapr/issues/4236 for next steps.

PR #211, opened 2022-03-14
Added fossa workflow to js-sdk

amulyavarote:
Signed-off-by: Amulya Varote <amulyavarote@QTM-SWATHIKIL-1.redmond.corp.microsoft.com>

# Description

Added a separate fossa workflow to js-sdk.

## Issue reference

Please reference the issue this PR will close: #201 

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests

### DIFF
--- a/.github/workflows/fossa.yaml
+++ b/.github/workflows/fossa.yaml
@@ -1,0 +1,28 @@
+name: fossa
+on:
+  repository_dispatch:
+    types: [fossa]
+  push:
+    branches:
+      - master
+      - release-*
+      - feature/*
+    tags:
+      - v*
+  pull_request:
+    branches:
+      - master
+      - release-*
+      - feature/*
+  workflow_dispatch: {}
+jobs:
+  fossa-scan:
+    runs-on: ubuntu-latest
+    steps:
+      - name: "Checkout code"
+        uses: actions/checkout@v3
+
+      - name: "Run FOSSA Scan"
+        uses: fossas/fossa-action@main # Use a specific version if locking is preferred
+        with:
+          api-key: b88e1f4287c3108c8751bf106fb46db6 # This is a PUSH ONLY token that is safe to be shared with the public


.github/workflows/fossa.yaml
```diff
@@ -0,0 +1,28 @@
...
+  fossa-scan:
+    runs-on: ubuntu-latest
+    steps:
+      - name: "Checkout code"
+        uses: actions/checkout@v3
+
+      - name: "Run FOSSA Scan"
+        uses: fossas/fossa-action@main # Use a specific version if locking is preferred
+        with:
+          api-key: b88e1f4287c3108c8751bf106fb46db6 # This is a PUSH ONLY token that is safe to be shared with the public
```
mukundansundar, line 28: Should FOSSA test also be run in the workflow? In all other repos that is the case.
cc @shubham1172

amulyavarote: This is included in the workflow only now if I am missing something here.

shubham1172: @amulyavarote there has to be two different steps for scan and test. Example: https://github.com/dapr/components-contrib/pull/1568

.github/workflows/fossa.yaml
```diff
@@ -0,0 +1,36 @@
...
+    env:
+      FOSSA_API_KEY: b88e1f4287c3108c8751bf106fb46db6 
+    steps:
+      - name: "Checkout code"
+        uses: actions/checkout@v3
+
+      - name: "Run FOSSA Scan"
+        uses: fossas/fossa-action@main # Use a specific version if locking is preferred
+        with:
+          api-key: ${{ env.FOSSA_API_KEY }} # This is a PUSH ONLY token that is safe to be shared with the public
```
shubham1172, line 30: nit: the comment can go to line 22

.github/workflows/fossa.yaml
```diff
@@ -0,0 +1,36 @@
...
+      - master
+      - release-*
+      - feature/*
+    tags:
+      - v*
+  pull_request:
+    branches:
+      - master
+      - release-*
+      - feature/*
```
shubham1172, line 16: nit: we can remove this as it's not used in this SDK

.github/workflows/fossa.yaml
```diff
@@ -0,0 +1,36 @@
...
+  workflow_dispatch: {}
+jobs:
+  fossa-scan:
```
shubham1172, line 19: We should add the condition
`if: github.repository_owner == 'dapr' # FOSSA is not intended to run on forks.`

Otherwise it will start running on forks and upload multiple copies of the repo to the FOSSA portal.

shubham1172 (changes requested):
Requesting some minor changes. We also need to add the FOSSA badge to README.

.github/workflows/fossa.yaml
shubham1172: `feature/*` branch

shubham1172 (approved):
lgtm, thanks @amulyavarote!

PR #248, opened 2022-04-16
Return a specific typed Promise<InvokeBindingResponseType> instead of Promise<object>

xAmast:
# Description

1. Pass the metadata object
2. Return a specific typed Promise<InvokeBindingResponseType> instead of Promise<object>
3. Support send binary data to binding

## Issue reference

Optimize the binding call without affecting the API, and there are no associated issues

## Checklist

* [x] Code compiles correctly
* [ ] Created/updated tests


XavierGeerinck:
Loving some of the ideas in this PR that will benefit a lot of the JS SDK going forward! Thanks a lot for this!

Some initial remarks: 
* Could you split up the PR in multiple ones? Since it fixes more than just the "Return specific Promise type" (I see metadata adding as well)
* Could you add a test?

We will review this as a team later today and come back to you.

src/implementation/Client/GRPCClient/binding.ts
```diff
@@ -12,8 +12,47 @@ limitations under the License.
...
+export class BindingRequest {
+  public readonly bindingName: string;
+  public readonly operation: string;
+  public data?: any;
+  public metadata: Record<string, string>;
+
+  constructor(bindingName: string, operation: string) {
+    this.bindingName = bindingName;
+    this.operation = operation;
+
```
shubham1172, line 32: nit: remove extra line

src/implementation/Client/GRPCClient/binding.ts
```diff
@@ -23,29 +62,65 @@ export default class GRPCClientBinding implements IClientBinding {
...
+    );
+  }
+
+  // Send an event to an external system
+  async send(
+    bindingName: string,
+    operation: string,
+    data: any,
+    metadata: Record<string, string>,
+  ): Promise<BindingResponse> {
```
shubham1172, line 120: Should it be this?
```suggestion
  ): Promise<InvokeBindingResponse> {
```

src/implementation/Client/GRPCClient/binding.ts
```diff
@@ -12,8 +12,47 @@ limitations under the License.
...
-import { InvokeBindingRequest, InvokeBindingResponse } from '../../../proto/dapr/proto/runtime/v1/dapr_pb';
+import {
+  InvokeBindingRequest as AutogeneratedInvokeBindingRequest,
+  InvokeBindingResponse as AutogeneratedInvokeBindingResponse,
+} from '../../../proto/dapr/proto/runtime/v1/dapr_pb';
 import IClientBinding from '../../../interfaces/Client/IClientBinding';
+import { types } from 'util';
+import { InvokeBindingResponse } from '../../../types/binding/InvokeBindingResponse.type';
+
+export class BindingRequest {
```
shubham1172, line 23: Do we need to export `BindingRequest` and `BindingResponse`?

Can we use [parameter properties](https://www.typescriptlang.org/docs/handbook/2/classes.html#parameter-properties) instead of members? Class definition will be cleaner.

shubham1172: Well, do we need these classes at all? `send` function can directly call `makeRequest` (instead of going via `invoke`, and `BindingRequest` or `BindingResponse` won't be needed.

xAmast: The reason for this is that we do not want to make destructive changes to the `send` function. The long-term plan is to add `invoke(req: BindingRequest): Promise<BindingResponse>` and `invoke(name: string, operation: string, data: any, metadata: Record<string, string>):   Promise<BindingResponse>`

xAmast:
I tried to add tests for metadata, but I didn't find a way to test metadata. I need help

src/implementation/Client/GRPCClient/binding.ts
shubham1172: `invoke` isn't a part of the public API, right? `IClientBinding` only has a `send`. Can you please share why do we need to add `invoke` and its two overloads?

shubham1172: Are you talking about changing the API name from `client.binding.send` to `client.binding.invoke`? https://docs.dapr.io/reference/api/bindings_api/#invoking-output-bindings we technically need only one method in our `IClientBinding`, which we have as `send` today. 

Do we really need the `BindingRequest` class? Isn't it simpler to use the params `name`, `operation`, `data`, and `metadata` directly? It makes sense from a consumer point of view and eliminates the need to pack and unpack the data. The class `BindingRequest` itself provides no value (it's more like a struct).

Similar argument for `BindingResponse`. Since we already have `InvokeBindingResponse`, do we need to introduce yet another class here?

/cc @XavierGeerinck

shubham1172:
@xAmast You can try [http bindings](https://docs.dapr.io/reference/components-reference/supported-bindings/http/) to test metadata.

XavierGeerinck:
> I tried to add tests for metadata, but I didn't find a way to test metadata. I need help

This one you can leave out for now seeing we are working on structuring this part better

src/implementation/Client/GRPCClient/binding.ts
XavierGeerinck: Agreed, deprecating `send` in favor of `invoke` is not something we want to do at this moment.

Regarding classes, I personally also would refrain from adding extra abstraction classes. The gRPC Classes build from dapr/dapr already provide a DTO, this would just add another DTO on top.

I can however see where this is coming from as the Java SDK does it through this way.

shubham1172:
ping @xAmast

XavierGeerinck:
We decided to split up this PR:
* Metadata resolution: issue #266 will be used for tracking
* Typing more of the code base (including a T prefix for the types): issue #279

Issue #218, opened 2022-03-22
Support customized header for invoker


Halomaple:
In many cases we need passing some params like `X-UserId` from web to backend in HTTP header, but current implementation of js-sdk is **blocking** the header. 
See `src/implementation/Client/HTTPClient/invoker.ts` line 31-36: 
```js
    if (method !== HttpMethod.GET) {
      // @ts-ignore
      fetchOptions.headers = {
        'Content-Type': 'application/json'
      };
    }
```

So I wonder whether we can add new param named header in function `invoke` to support this? 
Like `async invoke(appId: string, methodName: string, method: HttpMethod = HttpMethod.GET, data: object = {}, header = {}): Promise<object> {}`

shubham1172:
I support this request. It is something already supported by dotnet-sdk https://github.com/dapr/dotnet-sdk/issues/396

pnocera:
I did a fork and implemented it. That is 6 lines of code....could submit a PR eventually.

shubham1172:
@pnocera that's great, feel free to raise a PR. Thank you for helping out!

PR #257, opened 2022-05-04
added custom headers to invokers

pnocera:
# Description

Added custom headers parameter to invoke method.

## Issue reference

#218

## Checklist

* [ x ] Code compiles correctly
* [ ] Created/updated tests


XavierGeerinck (changes requested):
Would love to see tests added for this

Also, @shubham1172 would love your opinion on the header signature. Headers are passed after the data object, but I would like to see it earlier. What do you think of changing the signature to have an adaptable params object that merges those 2? e.g. params: { data: {}, headers: {} }

shubham1172:
Hi @pnocera, thank you for contributions! Can you please fix the [DCO](https://github.com/dapr/js-sdk/pull/257/checks?check_run_id=6288897998)? Can you please add tests too?

pnocera:
Sorry I don't have time to write the tests. I saw there aren't any for the invoke methods I could get "inspired" from. I tested it in our dev env with 6 microservices running in k8s without issues though.

shubham1172:
> Also, @shubham1172 would love your opinion on the header signature. Headers are passed after the data object, but I would like to see it earlier. What do you think of changing the signature to have an adaptable params object that merges those 2? e.g. params: { data: {}, headers: {} }

Agreed @XavierGeerinck. It will be better in terms of extensibility in the future. 

> Sorry I don't have time to write the tests. I saw there aren't any for the invoke methods I could get "inspired" from. I tested it in our dev env with 6 microservices running in k8s without issues though.

@pnocera You can refer this for invoke e2e tests: https://github.com/dapr/js-sdk/blob/915260bf2399999fe82307b31e5c70c1dc42c3ef/test/e2e/main.http.test.ts#L114

The "mock" listener should be handling the headers that you pass in to invoke. If you don't have bandwidth, I will create another issue and add the tests, let us know.

shubham1172 (changes requested):
LGTM overall, few changes requested.

XavierGeerinck:
Closing this as no reply from the owner and we will be tackling some deprecation for easier API access in 3.0

Issue #236, opened 2022-03-29
documentation: overhaul README.md


XavierGeerinck:
## Describe the proposal

It appears that the documentation is still a bit confusing. I would propose to overhaul the README in the following way:

* General introduction of what the Dapr JS SDK is
* Quick getting started example
* Link to the main docs.dapr.io page and steering people there
    * Maybe we can add a table to the different sections that each link to docs.dapr.io?
    * I would also love to include "Dapr Integerations" e.g. how to integrate with Nest.js, ...
* Link to companies using the Dapr JS SDK? 
* Link to blog articles using the Dapr JS SDK?
* References

Thoughts?  

- [ ] Update the Dapr docs to point to the latest index.md file and have a good overview on docs.dapr.io

XavierGeerinck:
@shubham1172 @amulyavarote @tanvigour

XavierGeerinck:
Might be a good idea indeed :-) not entirely sure as it involves running docker containers so developers running these should ensure they don't have port blocks

johnnyreilly:
Tiny thing, but it might be an idea to have installation instructions? eg

`npm install dapr-client --save`

I say this as the name of the repo is `js-sdk` and I couldn't immediately work out what the named of the published package was

shubham1172:
Thanks for the feedback @johnnyreilly, it's being added as a part of #223

XavierGeerinck:
@johnnyreilly the first PR for this has now been pushed, feel free to check and help us ensure it is clear enough

PR #271, opened 2022-05-19
Introduce test-init to initialize test services

XavierGeerinck:
# Description

Introduce a `test-init.sh` file to initialize the services required to run the tests. This makes it easier to know what has to be started to run tests.

### DIFF
--- a/scripts/test-init.sh
+++ b/scripts/test-init.sh
@@ -1,0 +1,29 @@
+#!/usr/bin/env bash
+#
+# Copyright 2022 The Dapr Authors
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#     http://www.apache.org/licenses/LICENSE-2.0
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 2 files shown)


scripts/test-init.sh
```diff
@@ -0,0 +1,29 @@
...
+echo "Starting MQTT Broker"
+docker run -d --rm --name emqx -p 1883:1883 -p 8081:8081 -p 8083:8083 -p 8883:8883 -p 8084:8084 -p 18083:18083 emqx/emqx
+
+echo "Starting Rabbit MQ"
+docker run -d --rm --hostname my-rabbitmq --name my-rabbitmq \
+    -p 0.0.0.0:5672:5672 -p 0.0.0.0:15672:15672 \
+    rabbitmq:3-management
+
+echo "Starting MongoDB"
+docker run -d --rm --name mongodb -p 27017:27017 mongo
```
shubham1172, line 29: Since we are adding two new components, were e2e tests for these components failing earlier?

XavierGeerinck: These components are required to run the tests :) always have been, just not documented well

shubham1172: Should we also add that to a readme - how to run e2e tests, as a separate item?

XavierGeerinck: I was planning on including that in the new README

shubham1172: Ok, it's already here in https://github.com/dapr/js-sdk/issues/236

shubham1172 (approved):

Issue #217, opened 2022-03-21
Logging support in JS-SDK


shubham1172:
## Describe the proposal
Contd. from #100.

Currently all SDK logs are console.log-ed, for example:
https://github.com/dapr/js-sdk/blob/1e1a8cbbfc3ec455a5d3c90a9616ae594d2e7c5a/src/implementation/Server/HTTPServer/HTTPServer.ts#L106

This proposal is to add a way for SDK consumers to pass a logger interface implementation while initializing it. If nothing is provided, the default implementation can be to log everything to stdout/stderr. This will allow SDK consumers to implement their own logging and control the log output destination (console, file, stream, syslog, etc.), the level of logging (debug, verbose, info, error, etc.).

With this, we will also need to consider improving **what** we log, i.e. the content itself.

PR #276, opened 2022-05-30
[WIP] Use a logging module and make it extensible

shubham1172:
# Description

This PR introduces a logging module and makes the SDK use that, instead of console.log statements. It also introduces a configuration for SDK users to customize (1) how much data to log, depending on logging levels, and (2) implementation of the logger.

## Issue reference

Please reference the issue this PR will close: #217 

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests


src/types/logger/LogLevel.ts
```diff
@@ -0,0 +1,23 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+/**
+ * LogLevel sets the level of logging that should be used.
+ */
+ export enum LogLevel {
+    error = 0,
```
shubham1172, line 18: Error

src/actors/runtime/AbstractActor.ts
```diff
@@ -52,7 +57,7 @@ export default abstract class AbstractActor {
   constructor(daprClient: DaprClient, id: ActorId) {
     this.daprClient = daprClient;
     this.actorClient = new ActorClient(daprClient.getDaprHost(), daprClient.getDaprPort(), daprClient.getCommunicationProtocol(), daprClient.getOptions());
-
+    this.logger = new Logger(daprClient.getOptions().logger)
```
XavierGeerinck, line 60: I would include the component name and area in here

e.g. 

```javascript
private readonly logger = new Logger(MyService.name);
```

ref: https://docs.nestjs.com/techniques/logger#using-the-logger-for-application-logging

src/logger/ConsoleLoggerService.ts
```diff
@@ -0,0 +1,43 @@
...
+    verbose(message: any, ...optionalParams: any[]): void {
+        console.log(`${this.getTime()} VERBOSE ${message}`, ...optionalParams);
+    }
+
+    debug(message: any, ...optionalParams: any[]): void {
+        console.debug(`${this.getTime()} DEBUG ${message}`, ...optionalParams);
+    }
+
+    private getTime(): string {
+        return new Date().toISOString();
```
XavierGeerinck, line 41: does this work? shouldn't it be `(new Date()).toISOString();`?

XavierGeerinck (changes requested):
2 minor comments

src/actors/runtime/AbstractActor.ts
shubham1172: Hmm, we won't be able to pass around Logger object anymore. We will need to pass around Logger options instead and have each class instantiate their own Logger based on `LoggerOptions`.

shubham1172: This has been updated to include component/area during the initialization. As discussed, logger is no longer passed explicitly. HttpClient options exposes logger options that is reused.

XavierGeerinck:
Looking good! Can we also get the constructor params to be all optional maybe? Wdyt?

Issue #260, opened 2022-05-13
Add warning message to `dapr-client` builds


shubham1172:
## Proposal

Per #207, add deprecation warnings to the `dapr-client` package builds and direct the user to move to `@dapr/dapr`.

XavierGeerinck:
Linked to #261

shubham1172:
npm postinstall scripts have stopped echoing to the terminal, see https://github.com/npm/cli/issues/3647

Another option for this is to use [npm deprecate]https://docs.npmjs.com/cli/v8/commands/npm-deprecate). We will need something like this:

```bash
npm deprecate dapr-client@ "dapr-client is deprecated. Please use @dapr/dapr instead. For more information, see https://github.com/dapr/js-sdk/issues/259".
```

@XavierGeerinck do you want to do this?

XavierGeerinck:
I can't... I don't have access to the `dapr-client` package

shubham1172:
Hello @XavierGeerinck, the versioning feature request is now being tracked separately. The PR for warning users on using dapr-client package is up and will close this issue.

PR #313, opened 2022-07-05
Add warning message to dapr-client builds

shubham1172:
# Description

This PR introduces a prebuild script to generate a src/version.ts file with package name and package version. 
The package name is then used in `DaprClient` to check if `dapr-client` is being used, and it shows a warning if true.

Also fixes build workflows to use `npm run build` instead of directly calling `scripts/build.sh`.

## Issue reference

Please reference the issue this PR will close: #260

## Checklist

* [X] Code compiles correctly
* [ ] Created/updated tests


XavierGeerinck (changes requested):
Should src/version.ts be committed as it is being auto generated?

shubham1172:
> Should src/version.ts be committed as it is being auto generated?

Thought about the same but it's needed by our build systems that do not run a `npm run build`. One option is to explicitly run it. What do you think?

shubham1172:
> Should src/version.ts be committed as it is being auto generated?

This has been checked-out of version control and workflows are updated. 

Only caveat, when users clone the repo for the first time, VS code will complain for the error until they run an `npm run build`.

XavierGeerinck (approved):

Issue #302, opened 2022-06-17
Add metadata to save field


XavierGeerinck:
## Expected Behavior

Should be able to use save with metadata

## Actual Behavior

We only allow key and value so cannot do this

## Steps to Reproduce the Problem

```javascript
    await daprClient.state.save(Config?.dapr?.components?.state?.cosmos?.name, [
      {
        key: "test",
        value: { hello: "world", partitionID: 2083 },
        metadata: {
          partitionKey: "partitionID"
        }
      }
    ])
```

Will result in

```
== APP == Error: {"errorCode":"ERR_STATE_SAVE","message":"failed saving state in state store state-azure-cosmosdb: BadRequest, Message: {\"Errors\":[\"PartitionKey extracted from document doesn't match the one specified in the header. Learn more: https:\\/\\/aka.ms\\/CosmosDB\\/sql\\/errors\\/wrong-pk-value\"]}\r\nActivityId
```

shubham1172:
@XavierGeerinck this has worked for me with transactions. https://github.com/shubham1172/dapr-cosmos-example/blob/reproduce-issue-1209/src/index.ts

Can you please share the partition key from the Cosmos DB portal on Azure?

XavierGeerinck:
![image](https://user-images.githubusercontent.com/591282/174455257-f4e70c96-a42a-4446-9927-4c43e018f9c7.png)

XavierGeerinck:
Tracking the PartitionKey not working in https://github.com/dapr/components-contrib/issues/1794

For the JS SDK we should implement the metadata part (see fetch example in the issue above)

shubham1172:
Got it, thanks!

XavierGeerinck:
For follow-up: https://github.com/dapr/components-contrib/issues/1794

PR #323, opened 2022-07-18
feat: add Metadata field to the State Save operation

XavierGeerinck:
Signed-off-by: Xavier Geerinck <xavier.geerinck@gmail.com>

# Description

_Please explain the changes you've made_

## Issue reference

Please reference the issue this PR will close: #302 

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests


test/e2e/grpc/client.test.ts
```diff
@@ -178,6 +184,34 @@ describe('grpc/client', () => {
...
+          key: 'key-2',
+          value: 'value-2',
+        },
+        {
+          key: 'key-3',
+          value: 'value-3',
+        },
+      ]);
+
+      const res = await client.state.get('state-redis', 'key-1');
```
shubham1172, line 211: In a scenario where we discard the etag, options and metadata, this test will still continue to work, right? Can we change that? Or if that requires some other work or is out of scope for this PR, we can track that separately.

shubham1172 (approved):

test/e2e/grpc/client.test.ts
XavierGeerinck: There are other tests not using these, so it should normally?

Issue #249, opened 2022-04-19
Configuration API - Subscribe All Keys


XavierGeerinck:
To not block the release, we are opening this issue for a further milestone. The test is in place but commented out until the dapr/dapr upstream bug has been resolved / looked into

## Expected Behavior

https://github.com/dapr/dapr/issues/4529

PR #327, opened 2022-07-20
Add Configuration API test for subscribe all keys

shubham1172:

# Description

With the fix in https://github.com/dapr/dapr/issues/4529, the test for subscribe has been added.

## Issue reference

Please reference the issue this PR will close: #249 

## Checklist

* [ ] Code compiles correctly
* [x] Created/updated tests


scripts/test-init.sh
```diff
@@ -11,9 +11,10 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-echo "Starting Docker"
-sudo service docker start
-sleep 3
+if (! docker stats --no-stream ); then
```
shubham1172, line 14: This is a cross-platform way since `service` might not be present.

Issue #333, opened 2022-08-02
PubSub Subscribe HTTP API response is invalid


shubham1172:
## Expected Behavior

JS-SDK should return the correct response from the subscribe message handler. See https://docs.dapr.io/reference/api/pubsub_api/#provide-a-route-for-dapr-to-discover-topic-subscriptions for the documentation.

From the docs,

Status | Description
-- | --
SUCCESS | Message is processed successfully
RETRY | Message to be retried by Dapr
DROP | Warning is logged and message is dropped
Others | Error, message to be retried by Dapr

## Actual Behavior

We send `{success: true}` or `{success:false}` which is incorrect. Code snippet:
https://github.com/dapr/js-sdk/blob/7412918d5b664a03d3284505758a9036f339c078/src/implementation/Server/HTTPServer/pubsub.ts#L44-L54

## Steps to Reproduce the Problem

Subscribe to messages.

shubham1172:
Did some more research here and turns out that today, only the retry functionality is broken since `{success:false}` does not mean anything to Dapr. Both the responses from SDK are logged as success scenarios in Dapr runtime. In the code snippet below, `AppResponse` turns out to be "" after parsing the response, which results in a fall-through to the success case.

https://github.com/dapr/dapr/blob/91af1367f29d77f16f8acee1cfff53830bc42b63/pkg/runtime/runtime.go#L1709-L1725

PR #334, opened 2022-08-02
[WIP] Correct HTTP response status for subscription

DeepanshuA:

# Description

This intends to fix n/ack of http response status correctly during subscription, as per https://docs.dapr.io/reference/api/pubsub_api/#expected-http-response

## Issue reference

Please reference the issue this PR will close: #333

## Checklist

* [x] Code compiles correctly
* [ ] Created/updated tests

### DIFF
--- a/src/enum/HttpReponseStatusForSubscription.enum.ts
+++ b/src/enum/HttpReponseStatusForSubscription.enum.ts
@@ -1,0 +1,25 @@
+/*
+Copyright 2022 The Dapr Authors
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+    http://www.apache.org/licenses/LICENSE-2.0
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+/**
+ * Http Response Status to be sent back to dapr runtime as subscription response, as per 
+ * https://docs.dapr.io/reference/api/pubsub_api/#expected-http-response
+ */
+export enum HttpResponseStatusForSubscription {
+    SUCCESS = "SUCCESS",  // Message is processed successfully
+    RETRY = "RETRY", // Message to be retried by Dapr
+    DROP = "DROP", // Warning is logged and message is dropped
+    Others = "Others" // Error, message to be retried by Dapr
+  }
+  
+  export default HttpResponseStatusForSubscription;
--- a/src/enum/SubscribedMessageHttpResponse.enum.ts
+++ b/src/enum/SubscribedMessageHttpResponse.enum.ts
@@ -15,11 +15,10 @@ limitations under the License.
  * Http Response Status to be sent back to dapr runtime as subscription response, as per 
  * https://docs.dapr.io/reference/api/pubsub_api/#expected-http-response
  */
-export enum HttpResponseStatusForSubscription {
+export enum SubscribedMessageHttpResponse {
     SUCCESS = "SUCCESS",  // Message is processed successfully
     RETRY = "RETRY", // Message to be retried by Dapr
     DROP = "DROP", // Warning is logged and message is dropped
-    Others = "Others" // Error, message to be retried by Dapr
   }
   
-  export default HttpResponseStatusForSubscription;
+  export default SubscribedMessageHttpResponse;
--- a/src/implementation/Server/HTTPServer/pubsub.ts
+++ b/src/implementation/Server/HTTPServer/pubsub.ts
@@ -15,7 +15,7 @@ import { TypeDaprPubSubCallback } from '../../../types/DaprPubSubCallback.type';
 import IServerPubSub from '../../../interfaces/Server/IServerPubSub';
 import HTTPServer from './HTTPServer';
 import { Logger } from '../../../logger/Logger';
-import HttpResponseStatusForSubscription from '../../../enum/HttpReponseStatusForSubscription.enum';
+import SubscribedMessageHttpResponse from '../../../enum/SubscribedMessageHttpResponse.enum';
 
 // https://docs.dapr.io/reference/api/pubsub_api/
 export default class HTTPServerPubSub implements IServerPubSub {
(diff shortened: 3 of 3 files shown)


shubham1172:
Thank you @DeepanshuA, this was really fast!

Quick question - were you able to test the existing behaviour?
Also it would be great to add e2e tests for this if possible, or we can create an issue to track it separately.

shubham1172:
cc @XavierGeerinck

src/enum/HttpReponseStatusForSubscription.enum.ts
```diff
@@ -0,0 +1,25 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+/**
+ * Http Response Status to be sent back to dapr runtime as subscription response, as per 
+ * https://docs.dapr.io/reference/api/pubsub_api/#expected-http-response
+ */
+export enum HttpResponseStatusForSubscription {
```
shubham1172, line 18:
```suggestion
export enum SubscribeMessageHttpResponse {
```

src/enum/HttpReponseStatusForSubscription.enum.ts
```diff
@@ -0,0 +1,25 @@
...
+    SUCCESS = "SUCCESS",  // Message is processed successfully
+    RETRY = "RETRY", // Message to be retried by Dapr
+    DROP = "DROP", // Warning is logged and message is dropped
+    Others = "Others" // Error, message to be retried by Dapr
```
shubham1172, line 22: (1) "Others" is not a part of the contract, and (2) Drop/Others are not used by the SDK. Shall we remove both of them?

shubham1172 (changes requested):
/cc @XavierGeerinck

DeepanshuA:
> Thank you @DeepanshuA, this was really fast!
> 
> Quick question - were you able to test the existing behaviour? Also it would be great to add e2e tests for this if possible, or we can create an issue to track it separately.

Yeah, I was able to test the existing behavior and after this change. Earlier, it always landed in runtime with appResponse.Status being blank ("") https://github.com/dapr/dapr/blob/91af1367f29d77f16f8acee1cfff53830bc42b63/pkg/runtime/runtime.go#L1719. But, now it lands either with status: "SUCCESS" or status: "RETRY".

I have created a separate issue for adding e2e test: https://github.com/dapr/js-sdk/issues/335

src/enum/HttpReponseStatusForSubscription.enum.ts
DeepanshuA: Will update.

src/enum/HttpReponseStatusForSubscription.enum.ts
DeepanshuA: I had kept them to keep in sync as per the docs. And, `Others` seem like a bit off, so definitely can remove. I am not sure about DROP, I just kept it for maintaining the sanctity of having all options available.

src/implementation/Server/HTTPServer/pubsub.ts
```diff
@@ -46,12 +47,12 @@ export default class HTTPServerPubSub implements IServerPubSub {
         await cb(data);
       } catch (e) {
         this.logger.error(`subscribe failed: ${e}`);
-        return res.send({ success: false });
+        return res.send({ status: HttpResponseStatusForSubscription.RETRY });
```
XavierGeerinck, line 50: Can we add a test to ensure this does not keep occurring? If the subscribe keeps on failing, we would have to keep on retrying, which could trigger an overload. Maybe we need a back-off timer with a max count here? Looking for input from everyone here :) (@shubham1172)

XavierGeerinck (changes requested):

src/implementation/Server/HTTPServer/pubsub.ts
DeepanshuA: Something like an exponential back-off and retries till like some static/configurable number of times. Else, at end, passing it to DROP. But, won't that mean that we would have to track the traceID and keep a count that how many times we have re-tried it?

XavierGeerinck: Note: to clarify, this is when a message comes in and we want to process it locally in our application. The logic is to ensure we can correctly process the message. We should maybe rename the logging from `this.logger.error(`subscribe failed: ${e}`);` to `this.logger.error(`[route-${topic}]  Could not process the incoming message, retrying`);`

I think it could be done simpler, `this.server.getServer().post` will create a route on the HTTP Server, so we can in the `HTTPServerPubSub` create an object that counts for the specific route how many times it was retried and the corresponding backoff timer. Whenever we succeed we then reset this timer and counter.

shubham1172: > If the subscribe keeps on failing, we would have to keep on retrying, which could trigger an overload. Maybe we need a back-off timer with a max count here?


@XavierGeerinck, failing to ACK a message successfully will result in Dapr failing to ACK the message to the broker. Since the brokers already have backoffs, max retries, etc. in place, does the app need to worry about it? This also means making these options configurable and potential clashes with broker configuration. Regarding backoff, the consumer can only respond back with HTTP response with a delay, this would require us to hold HTTP connections.

Here is an example of these configurations in Redis - https://docs.dapr.io/reference/components-reference/supported-pubsub/setup-redis-pubsub/ 

> We should maybe rename the logging from this.logger.error(subscribe failed: ${e}); to this.logger.error([route-${topic}] Could not process the incoming message, retrying);

Agreed! Maybe "attempting retry" since the broker may or may not actually respond back.

src/implementation/Server/HTTPServer/pubsub.ts
```diff
@@ -46,12 +47,12 @@ export default class HTTPServerPubSub implements IServerPubSub {
         await cb(data);
       } catch (e) {
         this.logger.error(`subscribe failed: ${e}`);
```
shubham1172, line 49: TODO: change the message to "message handling failed" or something.

src/enum/SubscribedMessageHttpResponse.enum.ts
```diff
@@ -0,0 +1,24 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+/**
+ * Http Response Status to be sent back to dapr runtime as subscription response, as per 
+ * https://docs.dapr.io/reference/api/pubsub_api/#expected-http-response
+ */
+export enum SubscribedMessageHttpResponse {
```
shubham1172, line 18: Hey Deepanshu, small typo - 
```suggestion
export enum SubscribeMessageHttpResponse {
```
or this one, based on the proto (preferably this, sorry for not suggesting this earlier)
```suggestion
export enum TopicEventHttpResponse {
```

DeepanshuA: That is intentional actually. It made better understanding for me: SubscribedMessageHttpResponse instead of SubscribeMessageHttpResponse.

shubham1172: It does convey - message from "subscribed" event. Whereas we send a different response when you actually subscribe to a a route/topic, this is from the callback handler. But I am okay with this too.

shubham1172 (approved):
LGTM, thank you @DeepanshuA!

XavierGeerinck (approved):

Issue #320, opened 2022-07-12
DaprClient is not re-using http.Agent when using static getInstance


XavierGeerinck:
the ActorRuntime is calling DaprClient.create which is a static method and will return a created DaprClient object. However, when using `http` we need to re-use the connection each time rather than creating a new port to interface over

https://github.com/dapr/js-sdk/blob/master/src/actors/runtime/ActorRuntime.ts#L58

Test should thus check when we create a DaprClient, and then a new one based on an existing one that we re-use the same client interface

https://github.com/dapr/js-sdk/blob/8bbfdbdcb536faaa338db9fd05bc73fd746c01b9/src/implementation/Client/DaprClient.ts#L139

PR #340, opened 2022-08-21
Reusing same http client instance

DeepanshuA:

# Description

Change to reuse same HTTPClient instance for multiple invocations.
Basically, HTTPClient is acting as a Singleton here.

Tested in stand-alone mode with Actors. I tried to invoke 100000 actors.

**Before**: It led to multiple http client instances getting formed and this each time a new http agent being formed.
It resulted in "Temporary error when accepting new connections: accept tcp [::]:3500: accept4: too many open files"

**After**: 100000 actor invocations got successful, without the above issue.

Some potential discussions: 

1. Though,  #320 talks specifically about http. Do we also need to cater grpc in same way?
2. HTTPClient is being exported and the contract may change - as it makes HTTPClient constructor as private but rather requires getInstance() static method to be called. I thought of using if/else inside constructor but: a) It looked like a bad design b) With `new HTTPClient`, a user expects new instance rather than, same instance being returned behind-the-scenes. c) It also led to couple of errors with other variables (their formation being conditional)
We can think of letting HTTPClient constructor to remain non-private but it can lead to leakage.

## Issue reference

Please reference the issue this PR will close: #320 

## Checklist

* [ ] Code compiles correctly
* [ ] Created/updated tests


XavierGeerinck:
Amazing PR! 🤩 Could you fix the linking issues so the test pass? Then I can merge

DeepanshuA:
> Amazing PR! 🤩 Could you fix the linking issues so the test pass? Then I can merge

Thanks! Sure, will get them sorted.

XavierGeerinck:
LGTM @shubham1172 additional review possible?

XavierGeerinck:
Can we also add gRPC in this scope? It will require the same

src/implementation/Client/HTTPClient/HTTPClient.ts
```diff
@@ -34,8 +34,9 @@ export default class HTTPClient implements IClient {
 
   private readonly httpAgent;
   private readonly httpsAgent;
+  private static _instance: HTTPClient;
 
-  constructor(
+  private constructor(
```
shubham1172, line 39: As discussed offline, the `HttpClient` object should not be cached. Rather cache the actual HTTP client, `fetch` in this case. Caching the `HttpClient` will return the same instance for different host/port/options combination.

shubham1172, line 39: As discussed offline, the `HttpClient` object should not be cached. Rather cache the actual HTTP client, `fetch` in this case. Caching the `HttpClient` will return the same instance for different host/port/options combination.

src/actors/client/ActorClient/ActorClient.ts
```diff
@@ -41,7 +41,7 @@ export default class ActorClient {
       }
```
shubham1172, line 41: Does the gRPC client re-use the same instance internally? If not, can we cache it the same way as HTTP?

shubham1172 (changes requested):

amulyavarote:
Changes look good to me! Thanks @DeepanshuA

amulyavarote (approved):

src/implementation/Client/HTTPClient/HTTPClient.ts
DeepanshuA: Along with `fetch`, had to take `HttpAgent` also in consideration or else, same connection issue was happening.

DeepanshuA:
> Can we also add gRPC in this scope? It will require the same

As discussed offline: It doesn't look quite straight-forward to cache grpc client. Instead, we plan to cache DaprClient and current PR would address only http client. I have created a separate issue https://github.com/dapr/js-sdk/issues/349 to track the discussed proposal.

src/implementation/Client/HTTPClient/HTTPClient.ts
```diff
@@ -57,17 +56,27 @@ export default class HTTPClient implements IClient {
...
+      HTTPClient.client = fetch;
+    }
 
     // Add a custom agent so we can decide if we want to reuse connections or not
     // we use an agent so we can reuse an open connection, limiting handshake requirements
     // Note: when using an agent, we will encounter TCPWRAP since the connection doesn't get destroyed
     if (this.options.isKeepAlive) {
-      this.httpAgent = new http.Agent({ keepAlive: true, keepAliveMsecs: 30 * 1000 });
-      this.httpsAgent = new https.Agent({ keepAlive: true, keepAliveMsecs: 30 * 1000 });
+      if(!HTTPClient.httpAgent) {
```
shubham1172, line 67: Do we need to make httpsAgent static as well? It is being passed as params to the fetch client.

DeepanshuA: I tested without making it static - Issue didn't resolve.

DeepanshuA: One possible reason is: 
https://stackoverflow.com/a/64208818/18848251 - It seems Agent helps keep connections open and when required, they can be resued.
But, with new Agent every time, a new pool is allocated - so effectively no reuse.

shubham1172: Ok, Java SDK reference which is doing something similar - https://github.com/dapr/java-sdk/blob/master/sdk/src/main/java/io/dapr/client/DaprHttpBuilder.java

Nit: to simplify the multiple if statements.

shubham1172 (approved):
LGTM, just a minor comment left.

src/implementation/Client/HTTPClient/HTTPClient.ts
DeepanshuA: Improved if-else and tested with that.

shubham1172 (approved):

src/implementation/Client/HTTPClient/HTTPClient.ts
```diff
@@ -57,18 +56,22 @@ export default class HTTPClient implements IClient {
...
-      this.httpAgent = new http.Agent({ keepAlive: true, keepAliveMsecs: 30 * 1000 });
-      this.httpsAgent = new https.Agent({ keepAlive: true, keepAliveMsecs: 30 * 1000 });
-    } else {
-      this.httpAgent = new http.Agent();
-      this.httpsAgent = new https.Agent();
-    }
+    const keepAlive = this.options.isKeepAlive;
+    const keepAliveMsecs = 30 * 1000; // it is applicable only when keepAlive is set to true
+
+      if(!HTTPClient.httpAgent) {
```
XavierGeerinck, line 69: Resolve correct indentation please, seems to be 4 spaces instead of 2?

XavierGeerinck (changes requested):
Small nit change regarding identation

src/implementation/Client/HTTPClient/HTTPClient.ts
DeepanshuA: My bad - corrected now.

shubham1172 (approved):

XavierGeerinck (approved):

Issue #283, opened 2022-06-01
Change configuration API response to dictionary


amulyavarote:
## Describe the proposal

Change the configuration API response to handle dictionary.
Main issue in dapr/dapr in pasted here.
https://github.com/dapr/dapr/issues/4556

PR #345, opened 2022-08-29
feat(configuration): Change configuration API response to dictionary

addjuarez:
# Description

Update configuration api to use dictionaries.

## Issue reference

Please reference the issue this PR will close: #283

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests


src/types/KeyConfig.type.ts
```diff
@@ -0,0 +1,18 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+import { ConfigurationItem } from "./configuration/ConfigurationItem";
+
+export type KeyConfigType = {
```
amulyavarote, line 16: I think we can change this to `ConfigurationType`

amulyavarote (changes requested):
Small comment to rename the type. The other changes look good to me.
And regarding the tests, can you please fix all the test cases and the github actions?

src/implementation/Client/GRPCClient/configuration.ts
```diff
@@ -52,17 +53,23 @@ export default class GRPCClientConfiguration implements IClientConfiguration {
...
-            key: item.getKey(),
-            value: item.getValue(),
-            version: item.getVersion(),
-            metadata: item.getMetadataMap().toObject().reduce((result: object, [key, value]) => {
-              // @ts-ignore
-              result[key] = value;
-              return result
-            }, {}),
-          }))
+        const configMap = Object()
```
shubham1172, line 56:
```suggestion
        const configMap: { [k: string]: ConfigurationItem } = {};
```

src/implementation/Client/GRPCClient/configuration.ts
```diff
@@ -111,18 +118,23 @@ export default class GRPCClientConfiguration implements IClientConfiguration {
 
     stream.on("data", async (data: SubscribeConfigurationResponse) => {
       streamId = data.getId();
+      const configMap = Object()
+      data.getItemsMap().forEach(function(v, k) {
+        const item: ConfigurationItem = {
+          key : k,
+          value : v.getValue(),
+          version : v.getVersion(),         
+          metadata :  v.getMetadataMap().toObject().reduce((result: object, [key, value]) => {
```
shubham1172, line 127: The conversion from gRPC metadata to KeyValuePair, shall we move this to Client.utils.ts and reuse the same in both places?

shubham1172:
This proto is a part of unreleased Dapr, right? We should wait for the next release before changing this here. Thoughts?

/cc @XavierGeerinck

shubham1172 (changes requested):

amulyavarote:
> This proto is a part of unreleased Dapr, right? We should wait for the next release before changing this here. Thoughts?
> 
> /cc @XavierGeerinck

Yes. Makes sense. We have to wait until the dapr changes goes in like we always do.

.github/workflows/test-e2e.yml
```diff
@@ -34,12 +34,12 @@ jobs:
...
     runs-on: ubuntu-latest
     env:
-      GOVER: 1.17
+      GOVER: 1.19
       DAPR_CLI_VER: 1.8.0
       DAPR_RUNTIME_VER: 1.8.0
       DAPR_INSTALL_URL: https://raw.githubusercontent.com/dapr/cli/master/install/install.sh
       DAPR_CLI_REF: '9472e6d977817243d6da28af3e3ca744414c54e6' # TODO, remove once DAPR CLI is upgraded to v1.7
-      DAPR_REF: ''
+      DAPR_REF: '683a436ff1367e4ff2d27da2d79069c04ec2c46d'
```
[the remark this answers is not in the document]
addjuarez: The changes made to the configuration api are breaking and require a newer version of dapr.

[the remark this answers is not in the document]
shubham1172: Yes, we should revert this and wait until Dapr's new version is released.

artursouza:
> > This proto is a part of unreleased Dapr, right? We should wait for the next release before changing this here. Thoughts?
> > /cc @XavierGeerinck
> 
> Yes. Makes sense. We have to wait until the dapr change goes in like we always do.

The Java SDK does not have this requirement as we allow to point to a git commit for dapr and build it as part of the SDK workflow for testing. Can this be achieved in this SDK too?

shubham1172:
> > > This proto is a part of unreleased Dapr, right? We should wait for the next release before changing this here. Thoughts?
> > > /cc @XavierGeerinck
> > 
> > 
> > Yes. Makes sense. We have to wait until the dapr change goes in like we always do.
> 
> The Java SDK does not have this requirement as we allow to point to a git commit for dapr and build it as part of the SDK workflow for testing. Can this be achieved in this SDK too?

Yes, we can do that @artursouza, but I only have one concern. JavaScript SDK has a separate release cadence than runtime/Java-SDK, this will cause unreleased Dapr proto definitions to be a part of JavaScript SDK releases. For Java it must be ok since they only cut their release after Dapr release gets cut. WDYT?

+ @XavierGeerinck for thoughts

XavierGeerinck:
I think version management is always going to be difficult as we need to couple releases to certain Dapr versions, thus subsequent updates will not receive certain bug fixes or features.

We could adapt the testing workflow to point to a certain Dapr commit version, but I don't see the real benefit for it, what would this be? As we would be testing against the newest and be ready, but in the end the PRs are there and validated by the developers at the end stage to be released when the latest Dapr version is.

shubham1172:
Hello @addjuarez, can you please fix conflicts?

shubham1172:
@addjuarez the build is failing due to prettify, you can run `npm run pretty-fix` to solve these issues.

shubham1172 (dismissed):
LGTM, thanks @addjuarez! Please fix the e2e tests.

.github/workflows/test-e2e.yml
```diff
@@ -34,12 +34,12 @@ jobs:
...
     runs-on: ubuntu-latest
     env:
-      GOVER: 1.17
-      DAPR_CLI_VER: 1.8.0
-      DAPR_RUNTIME_VER: 1.8.0
+      GOVER: 1.19
+      DAPR_CLI_VER: 1.9.1
+      DAPR_RUNTIME_VER: 1.9.0
       DAPR_INSTALL_URL: https://raw.githubusercontent.com/dapr/cli/master/install/install.sh
-      DAPR_CLI_REF: '9472e6d977817243d6da28af3e3ca744414c54e6' # TODO, remove once DAPR CLI is upgraded to v1.7
```
shubham1172, line 41: @XavierGeerinck FYI, this reduced about 1.5 mins from our e2e pipeline 🥳

src/implementation/Client/GRPCClient/configuration.ts
```diff
@@ -135,21 +127,16 @@ export default class GRPCClientConfiguration implements IClientConfiguration {
 
     stream.on("data", async (data: SubscribeConfigurationResponse) => {
       streamId = data.getId();
+      const items = data.getItemsMap();
+
+      if (items.getLength() == 0) {
```
shubham1172, line 132: FYI, I added this in my recent commit. Apparently `stream.on` is returning two entries per change. The first one is empty, and second one contains actual data. Added this snippet to simply ignore the first empty callback.

DeepanshuA:
LGTM!

amulyavarote:
Looks good to me!

amulyavarote (approved):

PR #346, opened 2022-08-29
Allow actors to be configurable

### LINKED ISSUE #89: Add support for Actor Reentrancy
## Describe the proposal

Add support for actor reentrancy

Needs to check if Node.js requires this seeing it's single threaded nature?

### LINKED ISSUE #300: Allow configuring Actors
Today Actors are not configurable by users. We have [ActorRuntimeConfig](https://github.com/dapr/js-sdk/blob/master/src/actors/runtime/ActorRuntimeConfig.ts) which is instantiated with default values in ActorRuntime directly.

https://github.com/dapr/js-sdk/blob/5780786e89dfa168a0a5cbdea84ab7edd0fc8097/src/actors/runtime/ActorRuntime.ts#L30

There should be a way for consumers to alter this value optionally. This is how it's set in other SDKs today: https://docs.dapr.io/developing-applications/building-blocks/actors/howto-actors/#actor-runtime-configuration

I propose exposing this as part of the actors `init` method. So the API will look like this:

```ts
const server = new DaprServer(serverHost, serverPort, daprHost, daprPort);
const client = new DaprClient(daprHost, daprPort);

await server.actor.init({ drainRebalancedActors: false }); // this is where config can be passed
server.actor.registerActor(MyActor);
await server.start();
```

shubham1172:
# Description

Actors was not configurable earlier, and the default runtime configuration was always being used. This PR allows actors to be configured via Dapr client's options. It introduces a new options type for actor. 

This change also moved away from Temporal to use plain strings. Temporal is an experimental spec and is not advised for production scenarios, see https://tc39.es/proposal-temporal/docs/

This change includes reentrancy configuration, but keeps per-actor type configuration for later (tracked via #243).

## Issue reference

Please reference the issue this PR will close #300 and close #89.

## Checklist

* [X] Code compiles correctly
* [X] Created/updated tests
* [X] Extended the documentation


shubham1172:
TODO: E2E test for config handler and reentrancy E2E flow.

src/types/actors/ActorReentrancyOptions.ts
```diff
@@ -0,0 +1,29 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+/**
+ * ActorReentrancyOptions provides options for configuring reentrancy in actors.
+ * For more, see https://docs.dapr.io/developing-applications/building-blocks/actors/actor-reentrancy/.
+ */
+export type ActorReentrancyOptions = {
```
DeepanshuA, line 18: Should we have ActorReentrancyOptions type present in ActorRuntimeOptions.ts itself? As `ReentrancyOptions` will be mainly used by `RuntimeOptions`.

test/e2e/http/actors.test.ts
```diff
@@ -79,6 +90,23 @@ describe('http/actors', () => {
...
+      expect(res.status).toBe(200);
+
+      const config = JSON.parse(await res.text())
+
+      expect(config.entities.length).toBe(8);      
+      expect(config.actorIdleTimeout).toBe('1h');
+      expect(config.actorScanInterval).toBe('30s');
+      expect(config.drainOngoingCallTimeout).toBe('1m');
+      expect(config.drainRebalancedActors).toBe(true);
+      expect(config.reentrancy.enabled).toBe(true);
```
DeepanshuA, line 105: Should we have one test, where-in if we don't set few options (like we don't set `reentrance.enabled` to be `true`), then we are able to test their default values getting set?

daprdocs/content/en/js-sdk-docs/js-actors/_index.md
```diff
@@ -64,6 +64,36 @@ export default class ParkingSensorImpl extends AbstractActor implements ParkingS
 }
 ```
 
+### Configuring Actor Runtime
+
+To configure actor runtime, use the `DaprClientOptions`. The various parameters and their default values are documented at [How-to: Use virtual actors in Dapr](https://docs.dapr.io/developing-applications/building-blocks/actors/howto-actors/#configuration-parameters).
+
```
DeepanshuA, line 70: Though we have mentioned in `ActorRuntimeOptions` that interval/timeout options use https://pkg.go.dev/time#ParseDuration format, can we specify explicitly the format to be used by duration kind of things? It may be handy for the user.

src/types/actors/ActorReentrancyOptions.ts
shubham1172: Are you talking about moving the type to that file, or embedding these properties directly in `ActorRuntimeOptions`? If the former, it's just a cosmetic change and maybe keeping in a separate file is cleaner. For the latter, it aligns to the JSON model required by runtime.

test/e2e/http/actors.test.ts
shubham1172: Default values won't be set from SDK. Runtime is responsible for that. I can add a test that confirms these values are omitted - but it will be the same as the existing unit test for that method.

daprdocs/content/en/js-sdk-docs/js-actors/_index.md
shubham1172: Yeah, it maybe helpful, just wondering if it will clutter the comment too much, and then if something changes in the `time.ParseDuration` our comment will be stale. WDYT?

src/utils/Actors.util.ts
```diff
@@ -0,0 +1,41 @@
...
+limitations under the License.
+*/
+
+import { ActorRuntimeOptions } from "../types/actors/ActorRuntimeOptions";
+
+/**
+ * Get response object for /dapr/config API call.
+ * See https://docs.dapr.io/reference/api/actors_api/#get-registered-actors
+ */
+export function getRegisteredActorResponse(entities: string[], options: ActorRuntimeOptions): {[k: string]: any} {
```
XavierGeerinck, line 20: This method feels a bit bloated / complex for what it should do I think? Isn't there a way that we can create the response more easily? (e.g. by specifying which keys to copy and just copy them over if they are set?)

Other than that general remarks if we can't simplify it:
* Please no single line ifs
* Let's use our KeyValueType if possible
* Let's maybe create a new type for this specifically?

XavierGeerinck (changes requested):

src/types/actors/ActorReentrancyOptions.ts
DeepanshuA: Yeah, the former. To me it seemed no need of an extra file and have this type reside in `RuntimeOptions` file. But, I am fine with keeping it separate as well.

test/e2e/http/actors.test.ts
DeepanshuA: Yups, defaut values are set in runtime only. As it is an E2E test, so, I meant if we can somehow fetch Actors' configuration and test the expectations.

daprdocs/content/en/js-sdk-docs/js-actors/_index.md
DeepanshuA: I think it would be definitely helpful. As an example, when I was very new in Dapr, I was trying to set some duration property in a yaml and I had to search at some places/fail few times due to that property before I could determine the correct syntax. For any new user, it should be right there wherever we are documenting. 
I don't think `time.ParseDuration` syntax would change as such. In general, a better way for any duration filed can be if as Dapr are able to define the syntax that we recommend at a single place and any/all duration document can refer that.

src/utils/Actors.util.ts
shubham1172: >by specifying which keys to copy and just copy them over if they are set? 


One of the motivation was to not have the two directly coupled, i.e., `options.XYZ` can be mapped to `options.actorXYZ`. Since we are not doing it today, i.e. we have 1:1 mapping, let me update this.

daprdocs/content/en/js-sdk-docs/js-actors/_index.md
shubham1172: > In general, a better way for any duration filed can be if as Dapr are able to define the syntax that we recommend at a single place and any/all duration document can refer that.

Agreed. This should be locally updated in runtime documentation, e.g., https://docs.dapr.io/developing-applications/building-blocks/actors/howto-actors/#actor-timers-and-reminders, and we can refer that.

shubham1172: https://github.com/dapr/docs/issues/2766 created to track.

src/utils/Actors.util.ts
shubham1172: We can technically remove the util also now, just that the unit test cases are a definite way to know we didn't mess up the de-structuring and adding.

XavierGeerinck (approved):

daprdocs/content/en/js-sdk-docs/js-actors/_index.md
DeepanshuA: Can we at least for now provide link to https://pkg.go.dev/time#ParseDuration ?

shubham1172: Oh yeah, sure! I was thinking about this src/types/actors/ActorRuntimeOptions.ts all this while, which has the link to it. Let me add it here, good catch!

shubham1172: Sure, it's present in src/types/actors/ActorRuntimeOptions.ts (for developers), adding it here.

XavierGeerinck (approved):

Issue #339, opened 2022-08-20
GRPCClient doesn't pass metadata to binding: unable to send data to Redis Output Binding


alecor191:
## Expected Behavior

When sending data to a Redis Output Binding using `DaprClient`, data is successfully accepted by Redis.

## Actual Behavior

Sending data to Redis Output Binding fails, as the provided `key` metadata value doesn't seem to get forwarded to Redis. As this is a mandatory metadata field, Redis binding fails with the following error:

```
Error: 13 INTERNAL: error when invoke output binding node-sub-app-blob: redis binding: missing key on write request metadata
```

## Steps to Reproduce the Problem

1. Have Dapr CLI installed
2. Create an Output Binding as follows:

  ```yaml
  apiVersion: dapr.io/v1alpha1
  kind: Component
  metadata:
    name: my-redis-binding
  spec:
    type: bindings.redis
    version: v1
    metadata:
      - name: redisHost
        value: localhost:6379
      - name: redisPassword
        value: ""
  ```

3. Create some simple NodeJS app and use the following code to send some data to the Redis binding:

  ```typescript
  const client = new DaprClient(daprHost, daprGrpcPort, CommunicationProtocolEnum.GRPC);
  
  await client.binding.send(
      "my-redis-binding",
      "create",
      "some message",
      { key: "my-key" });
  ```

4. The `send` request fails with the error message mentioned above.

## Additional information

- @dapr/dapr package version `2.3.1` used
- Dapr CLI version `1.8.1`
- I used Python SDK to do the same as above and it works fine.
- Briefly looking at `@dapr\dapr\implementation\Client\GRPCClient\binding.js` seems to give some hints, that the provided metadata values are not used (yet)? Notice `@todo` message as well as `_metadata` parameter not being used.

```typescript
    // Send an event to an external system
    // @todo: should pass the metadata object <<<===== NOT SUPPORTED YET?
    // @todo: should return a specific typed Promise<TypeBindingResponse> instead of Promise<object>
    send(bindingName, operation, data, _metadata = {}) {  <<<===== _metadata not used?
        return __awaiter(this, void 0, void 0, function* () {
            const msgService = new dapr_pb_1.InvokeBindingRequest();
            msgService.setName(bindingName);
            msgService.setOperation(operation);
            msgService.setData(SerializerUtil.serializeGrpc(data).serializedData);
            const client = yield this.client.getClient();
            return new Promise((resolve, reject) => {
[...]

PR #348, opened 2022-09-01
feat(bindings): Pass Metadata in grpc binding invocation

DeepanshuA:

# Description

Adds metadata capability to output binding invocation.

## Issue reference

Please reference the issue this PR will close: #339 

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests


DeepanshuA:
E2E test is failing here, since we are using Dapr 1.8.0, the latest released major version (https://github.com/dapr/js-sdk/blob/c3f6347f632a508944f7b8296298f665f32a9d6f/.github/workflows/test-e2e.yml#L39)
But, it needs change from https://github.com/dapr/dapr/pull/5049, specifically https://github.com/dapr/dapr/blob/490ef74dbd06e13a4557a249afd58d1abecb01af/pkg/grpc/api.go#L547

DeepanshuA:
Commented out E2E test for now - E2E tests should pass now.
Would un-comment after dapr 1.9 release.

test/components/binding-redis.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: dapr.io/v1alpha1
```
shubham1172, line 1: Can't we reuse the existing MQTT binding?

shubham1172:
> E2E test is failing here, since we are using Dapr 1.8.0, the latest released major version (
> 
> https://github.com/dapr/js-sdk/blob/c3f6347f632a508944f7b8296298f665f32a9d6f/.github/workflows/test-e2e.yml#L39
> 
> )
> But, it needs change from [dapr/dapr#5049](https://github.com/dapr/dapr/pull/5049), specifically https://github.com/dapr/dapr/blob/490ef74dbd06e13a4557a249afd58d1abecb01af/pkg/grpc/api.go#L547

@DeepanshuA is it possible to use some other metadata for testing? i.e. if you consider MQTT, we can override the topic via metadata ([docs](https://docs.dapr.io/reference/components-reference/supported-bindings/mqtt/#set-topic-per-request)) and test it.

test/components/binding-redis.yaml
XavierGeerinck: Let's add it for Redis specifically as well maybe as it's a common component?

shubham1172: Sure, shall we remove MQTT then too? It will ease setup. Redis is more commonly available.

XavierGeerinck: I would keep MQTT as we want to test on different platforms, no? MQTT is used quite broadly btw :)

DeepanshuA: To me it seems fine to have redis and mqtt both.
I used redis here specifically, as for redis metadata "key" is crucial (https://github.com/dapr/components-contrib/blob/b18e73d028e1b6553973efd449085a88d0bb9690/bindings/redis/redis.go#L73) or else this output binding won't serve any purpose.

shubham1172: Ok, I would suggest having one working e2e test (from any component for what it's worth), and then uncomment the other one later.

shubham1172 (dismissed):

shubham1172 (approved):

Issue #330, opened 2022-07-24
can't send msg >4mb


giang12:
## Expected Behavior
to be able to invoke/pub with req size > 4mb fixed...for both http and grpc clients
## Actual Behavior
 grpc client:  error "code": 8,
    "details": "Received message larger than max (6632286 vs. 4194304)",
http client: nothing happen.. return undefined...

## Steps to Reproduce the Problem
try invoke remote services via sidecar and see...
seem like missing 3rd options params for grpc client https://github.com/dapr/js-sdk/blob/7bf1bb8e965fee48144e6f9b5707f39e6cf4950b/src/implementation/Client/GRPCClient/GRPCClient.ts#L76

giang12:
try setting 
grpc.max_send_message_length
grpc.max_receive_message_length
from https://www.npmjs.com/package/@grpc/grpc-js..as gathered from https://github.com/dapr/js-sdk/issues/70
still no success..

XavierGeerinck:
Can you list your Dapr run command?

giang12:
`dapr run --app-id ingress-app --app-port 8080 --dapr-grpc-port 3500 --dapr-http-max-request-size 20 -- node bin/cli.js`
(there is no equivalent --dapr-hrpc-max-request-size?)
tx for getting back so quickly

giang12:
https://github.com/alvinhenrick/go-sdk/blob/52e8721128799f3ad77136c6d82e4f769692be26/examples/service/custom-grpc-client/main.go#L29 i saw in go-sdk they recommend manually setting up grpc conns https://github.com/dapr/go-sdk/pull/197

shubham1172:
Hello @giang12, is HTTP mode working with `--dapr-http-max-request-size`?

giang12:
hi @shubham1172 no unfortunately.. just tried again to make sure. I ran both sender and receiver in http mode like this 
dapr run --app-id ingress-app --app-port 8080 --dapr-http-port 3500 --dapr-http-max-request-size 44 -- node bin/cli.js
`dapr run --app-id echo-app --app-protocol http  --app-port 50051 --dapr-http-port 3503 --dapr-http-max-request-size 44 -- node dist/index.js`
![image](https://user-images.githubusercontent.com/5290460/180907184-e95ad814-ec38-4bff-bc94-5d28b9400116.png)
never reached 2nd console.log..just return without any error somehow, just an empty `{}`

shubham1172:
[...]

PR #356, opened 2022-09-06
fix: Can not request data larger than 4MB with grpc (#330)

JY29:
# Description

There is no way to change the grpc default option before.So when respond is larger than 4MB this error occured. I fix this bug by adding grpc options to dapr options

## Issue reference
#330 

## Checklist

* [x] Code compiles correctly
* [x] Created/updated tests
* [x] Extended the documentation


XavierGeerinck:
Hi! Thanks a lot for your PR, amazing to see the nice contribution.

I reviewed your PR and while I think it tackles the issue, we were aiming towards a more general approach for the entire DaprClient side of things.

What we were thinking of is to solve this at the `DaprClient` side of things. Here we push `DaprClientOptions` where we would like to add something such as `maxBodySizeMb: number` that propagates to both HTTP and gRPC (we have 2 protocols, so both should be adapted for it)

https://github.com/dapr/js-sdk/blob/22fa4077bd7295c6158c90c241c7f79351d9a038/src/implementation/Client/DaprClient.ts#L85

https://github.com/dapr/js-sdk/blob/22fa4077bd7295c6158c90c241c7f79351d9a038/src/types/DaprClientOptions.ts#L16

This way we can configure both `HTTPServer` and `GRPCServer` and abstract the logic for the end consumer

https://github.com/dapr/js-sdk/blob/22fa4077bd7295c6158c90c241c7f79351d9a038/src/implementation/Server/HTTPServer/HTTPServer.ts#L45

https://github.com/dapr/js-sdk/blob/22fa4077bd7295c6158c90c241c7f79351d9a038/src/implementation/Server/GRPCServer/GRPCServer.ts#L45

Putting all of this together would result in the DaprClient that could be instantiated like this:

```javascript
const client = new DaprClient(daprHost, daprPort, CommunicationProtocolEnum.HTTP, {
    maxBodySizeMb: 10
});
```

Is this something you could help with?

P.S. We also can't merge a PR without the tests, so please add those as well

JY29:
thx, Ur idea seems better.I will try to write better and add some tests about this feature.

shubham1172:
Closing in favor of #381

Issue #282, opened 2022-05-31
Add APP_API_TOKEN


XavierGeerinck:
## Describe the proposal

https://github.com/dapr/docs/blob/0ce74a16c2cea299d9b841f21aa2fb276732d081/daprdocs/content/en/operations/security/app-api-token.md

shubham1172:
@XavierGeerinck shall we reopen this?

XavierGeerinck:
> @XavierGeerinck shall we reopen this?

Yes, will close the other

shubham1172:
👋🏻

PR #364, opened 2022-09-15
feat(client): add dapr api token to headers

ahmedhesham6:
# Description

Added `Dapr Api Token` to client options and http headers

## Issue reference

Please reference the issue this PR will close: #282

## Checklist

* [X] Code compiles correctly
* [X] Created/updated tests


src/types/DaprClientOptions.ts
```diff
@@ -30,4 +30,9 @@ export type DaprClientOptions = {
    * Options related to actors.
    */
   actor?: ActorRuntimeOptions;
+
+  /**
+   * Dapr Api Token
+   */
+  daprApiToken?: string;
```
shubham1172, line 37:
```suggestion
  /**
   * API token to authenticate with Dapr.
   * See https://docs.dapr.io/operations/security/api-token/.
   */
  daprApiToken?: string;
```

src/implementation/Client/HTTPClient/HTTPClient.ts
```diff
@@ -151,6 +151,12 @@ export default class HTTPClient implements IClient {
 
     if (!params?.headers) {
       params.headers = {};
+      if(this.options.daprApiToken){
```
shubham1172, line 154:
```suggestion
      if (this.options.daprApiToken) {
```

src/implementation/Client/HTTPClient/HTTPClient.ts
```diff
@@ -151,6 +151,12 @@ export default class HTTPClient implements IClient {
 
     if (!params?.headers) {
```
shubham1172, line 152: This adds the API token to headers if they are missing, but if the caller has passed headers, we also need to append the token.

test/e2e/http/client.test.ts
```diff
@@ -24,9 +24,9 @@ describe('http/client', () => {
   // we put a timeout of 10s since it takes around 4s for Dapr to boot up
   beforeAll(async () => {
     client = new DaprClient(daprHost, daprPort, CommunicationProtocolEnum.HTTP, {
-      isKeepAlive: false
+      isKeepAlive: false,
+      daprApiToken: 'token',
```
shubham1172, line 28:
```suggestion
      daprApiToken: 'securedtoken',
```

Note, for the tests to work you also need to edit package.json and add `DAPR_API_TOKEN` environment variable, similar to `TEST_SECRET_1`

shubham1172 (changes requested):
Thanks for your PR @ahmedhesham6 🥳 Left a few comments!

src/implementation/Client/HTTPClient/HTTPClient.ts
```diff
@@ -11,16 +11,16 @@ See the License for the specific language governing permissions and
...
+import http from 'http';
+import https from 'https';
+import { DaprClientOptions } from '../../../types/DaprClientOptions';
 import { Settings } from '../../../utils/Settings.util';
-import { THTTPExecuteParams } from "../../../types/http/THTTPExecuteParams.type"
-import { Logger } from "../../../logger/Logger";
-import HTTPClientSidecar from "./sidecar";
+import { THTTPExecuteParams } from '../../../types/http/THTTPExecuteParams.type';
+import { Logger } from '../../../logger/Logger';
+import HTTPClientSidecar from './sidecar';
```
shubham1172, line 23: Lot of similar changes in this PR, can you please revert them @ahmedhesham6? I am okay with these but we should track it separately + configure ESLint. 

Also looks like there are some extra commits in this PR.

ahmedhesham6: I don't know what is the issue with linting so I will reopen this pr

ahmedhesham6: Done please review.

shubham1172 (dismissed):
Thank you @ahmedhesham6, LGTM! 🚀

src/implementation/Client/HTTPClient/HTTPClient.ts
```diff
@@ -153,6 +153,13 @@ export default class HTTPClient implements IClient {
       params.headers = {};
     }
 
+    if (this.options.daprApiToken) {
```
XavierGeerinck, line 156: Please rework this to: 

```javascript
if (this.options.daprApiToken) {
    params.headers["dapr-api-token"] = this.options.daprApiToken;
}
```

I might even want to see: 

```javascript
params.headers["dapr-api-token"] = this.options.daprApiToken || undefined; 
```

but then I would like @shubham1172 his opinion

XavierGeerinck (changes requested):
Would love to see soe changes first for overal health of the repo:
1. Let's recode the if-check as below
2. The client is not being tested as it should. We are enforcing all tests to pass with the API token in there, which might make it harder if we want to test with other authentication methods. Instead, we should work by creating a new test that creates a new client that allows to authenticate so that we can testauthentication on success and on failure itself.

I can understand that 2 might sound difficult and long, but I would encourage everyone to look at it as it helps us in growing personally and stresses importance of stability in the long run :)

src/implementation/Client/HTTPClient/HTTPClient.ts
shubham1172: Agreed, first one looks cleaner (without `undefined`).

ahmedhesham6: I agree too.

test/unit/main/DaprClient.test.ts
```diff
@@ -40,4 +40,9 @@ describe('DaprClient', () => {
       expect(values[0]).toEqual(values[1]);
     });
   });
+
+  it('should return dapr api token', () => {
```
XavierGeerinck, line 44: Please create a new category named "Dapr API Token" with 2 test:
1. This one to make sure we can communicate and see the options
2. Another one showing failure

P.S. I approved test runs but I think this will make the other tests fail since they require an API token which is now missing

ahmedhesham6: How should it fail ?

XavierGeerinck: By creating a client without the API token that should provide a forbidden error

ahmedhesham6: But that will require creating a new client in E2E testing and invoking a request which will return Forbidden.
I feel like i am testing dapr itself not the client, the main purpose of this PR is just adding the token to the header so I guess a more suitable test for the use case is checking if the token is being sent in the headers or not.

XavierGeerinck: Agreed, will approve! Thanks a lot for your collaboration 😊

ahmedhesham6: I guess we needed to remove `DAPR_API_TOKEN=securedtoken` from package.json

Issue #296, opened 2022-06-11
When the invoker experience 404 it throws empty error


avifatal:
## Expected Behavior

The error needs to be more informative. its hard to detect on a large scale apps

## Actual Behavior

```
== APP == [Nest] 1585562  - 11/06/2022, 8:32:03   ERROR [ExceptionsHandler] ""
== APP == Error: ""
== APP ==     at HTTPClient.<anonymous> (/home/avi/work/tokenct/apps/api/node_modules/.pnpm/dapr-client@2.2.1/node_modules/dapr-client/implementation/Client/HTTPClient/HTTPClient.js:128:23)
== APP ==     at Generator.next (<anonymous>)
== APP ==     at fulfilled (/home/avi/work/tokenct/apps/api/node_modules/.pnpm/dapr-client@2.2.1/node_modules/dapr-client/implementation/Client/HTTPClient/HTTPClient.js:17:58)
== APP ==     at processTicksAndRejections (node:internal/process/task_queues:96:5)
```

## Steps to Reproduce the Problem

Just invoke an http call that is not answering on the other side

I think these lines should be more informative when throwing: https://github.com/dapr/js-sdk/blob/a5ae563deac37efef80ea3f801601a7cd8fa3837/src/implementation/Client/HTTPClient/HTTPClient.ts#L178

salmankhan-prs:
Hi @shubham1172  can you help to understand what  we need to do ? 
I think we need add extra message when status is 404 ,500 ,502

shubham1172:
> Hi @shubham1172 can you help to understand what we need to do ? I think we need add extra message when status is 404 ,500 ,502

Hi @salmankhan-prs, I think for starters we can respond with the status code and the error message, as proposed in the issue description,.

PR #415, opened 2022-10-22
fix when the invoker experience 404 it throws empty error 

salmankhan-prs:
# Description

Fix when the invoker experience 404 it throws empty error.

```js
    // 2XX -> OK; 3XX -> Redirects and Found
    if (res.status >= 200 && res.status <= 399) {
      return txtParsed;
    }

    // All the others
    else {
      this.logger.debug("Execute response text: %s", txtParsed);
      throw new Error(JSON.stringify({ status: res.status, originError: txtParsed }));
    }
```
All the  error status handled in one condition.

## Issue reference

Please reference the issue this PR will close: #296 

## Checklist

- [x] Code compiles correctly
- [ ] Created/updated tests

### DIFF
--- a/src/implementation/Client/HTTPClient/HTTPClient.ts
+++ b/src/implementation/Client/HTTPClient/HTTPClient.ts
@@ -206,8 +206,16 @@ export default class HTTPClient implements IClient {
 
     // All the others
     else {
-      this.logger.debug("Execute response text: %s", txtParsed);
-      throw new Error(JSON.stringify({ status: res.status, originError: txtParsed }));
+      this.logger.debug(`Execute response with  status: ${res.status} and text: ${txtParsed}`);
+      throw new Error(
+        JSON.stringify({
+          error: "UNKNOWN",
+          error_msg: `An unknown problem occurred and we got the status ${res.status} with response ${JSON.stringify(
+            res,
+          )}`,
+          status: res.status,
+        }),
+      );
     }
   }
 }
--- a/src/implementation/Client/HTTPClient/HTTPClient.ts
+++ b/src/implementation/Client/HTTPClient/HTTPClient.ts
@@ -203,29 +203,11 @@ export default class HTTPClient implements IClient {
     if (res.status >= 200 && res.status <= 399) {
       return txtParsed;
     }
-    // 400 = Bad Request, 401 = Unauthorized, 404 = Not Found
-    else if (res.status === 400 || res.status === 401 || res.status === 404) {
-      throw new Error(JSON.stringify(txtParsed));
-    }
-    // 403 = Forbidden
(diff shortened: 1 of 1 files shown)


salmankhan-prs:
@shubham1172   please review this PR

salmankhan-prs:
@shubham1172  @XavierGeerinck  please review the PR

src/implementation/Client/HTTPClient/HTTPClient.ts
```diff
@@ -203,29 +203,11 @@ export default class HTTPClient implements IClient {
...
-      throw new Error(JSON.stringify(txtParsed));
-    }
-    // 500 = Internal Server Error, 502 = Bad Gateway
-    else if (res.status === 500 || res.status === 502) {
-      throw new Error(JSON.stringify(txtParsed));
-    }
+
     // All the others
     else {
       this.logger.debug("Execute response text: %s", txtParsed);
```
shubham1172, line 209: Please add status code as well.

src/implementation/Client/HTTPClient/HTTPClient.ts
```diff
@@ -203,29 +203,11 @@ export default class HTTPClient implements IClient {
...
-      throw new Error(
```
shubham1172, line 221: @XavierGeerinck are we breaking the API contract with this PR? If so, do you think we should move it to 3.0.0?

XavierGeerinck:
I would use the keys `error` and `error_msg` with additional the `status to ensure we don't break the contract. This way we can merge before 3.0.0

shubham1172:
> I would use the keys `error` and `error_msg` with additional the `status to ensure we don't break the contract. This way we can merge before 3.0.0

This is better, @salmankhan-prs can you please incorporate it?

salmankhan-prs:
@shubham1172 you want me  to do something like this ?




  
    else {
      this.logger.debug(`Execute response with  status: ${res.status} and text: ${txtParsed}`);
      throw new Error(
        JSON.stringify({
          error: "UNKNOWN",
          error_msg: `An unknown problem occurred and we got the status ${res.status} with response ${JSON.stringify(
            res,
          )}`,
        }),
      );
      }

shubham1172:
@salmankhan-prs something like this:
```js
throw new Error(
        JSON.stringify({
          error: "UNKNOWN",
          error_msg: `An unknown problem occured and we got the status ${res.status} with response ${JSON.stringify(
            res,
          )}`,
          status: res.status
        }),
      );
```

`status` field is redundant but I am ok with it since we will eventually have a status and an error message.

salmankhan-prs:
@shubham1172  I have pushed the changes .please verify it

shubham1172 (approved):

Issue #409, opened 2022-10-09
Calling an actor method that doesnt exist on the class should not cause the app to crash


danyo1399:
## Expected Behavior
Calling an actor method that doesnt exist on the class should not cause the app to crash

## Actual Behavior
It does cause the app to crash

## Steps to Reproduce the Problem
Create an actor class in an app and start it using dapr runtime. call a nonexistent method on the actor type using postman causes the process to crash

danyo1399:
screenshot

https://share.gistdoc.tty.nz/shared/15bd48b27d2e8ce51e50ca8a6e60f166/dapr-js-sdk-calling-invalid-method-on-actor-crashes

PR #381, opened 2022-09-27
feat: Migrate from Restana to Express, add large data support and misc changes

XavierGeerinck:
# Description

This PR introduces
* \> 4Mb data support
* Correct data serialization and content-type header setting including tests for inflation
* Invocation is now possible for non-string or object data
* Removed uuid in favor of `crypto.randomUUID()`
* Changed IRequest and IResponse from interface to type. This means that we now don't export the type by default anymore and require an import as `{ IRequest }`
* Converted `IServerImplType` and `IServerType` to a type instead of an interface
* Possibility to add your own `express` instance through the `serverHttp` option

## Issue reference

We strive to have all PR being opened based on an issue, where the problem or feature have been discussed prior to implementation.

Please reference the issue this PR will close: #330, #440, #186 

## Checklist

Please make sure you've completed the relevant tasks for this PR, out of the following list:

* [x] Code compiles correctly
* [x] Created/updated tests
* [x] Extended the documentation

PR #422, opened 2022-11-15
Error shouldn't cause Client Fail If Actor Non-Existent method called

DeepanshuA:

# Description

Currently, for registered actors, if a non-existent method is called (let's say via postman), it makes App crash. This behavior is not correct and rather only an error with proper message needs to be sent.
This PR intends to fix this.

Got this response, after this change:

image

## Issue reference

Please reference the issue this PR will close: #409 

## Checklist

- [ ] Code compiles correctly
- [ ] Created/updated tests


DeepanshuA:
> @DeepanshuA is this ready for review?

Ok, All tests seem to be passing on this PR.
I just went ahead to cross-check that e2e test that I have included has also passed.
But, it seems actor tests are not running as a part of all these tests.

I am checking my test separately now.

XavierGeerinck:
Discussed: look at reverting the `package.json` back and see what the core issue is + wait on PR #381

src/implementation/Server/HTTPServer/actor.ts
```diff
@@ -98,13 +99,18 @@ export default class HTTPServerActor implements IServerActor {
...
+    try {
+      const result = await ActorRuntime.getInstance(this.client.getDaprClient()).invoke(
+        actorTypeName,
+        actorId,
+        methodName,
+        dataSerialized,
+      );
+      return this.handleResult(res, result);
+    } catch (err) {
+      res.statusCode = HttpStatusCode.BAD_REQUEST;
```
shubham1172, line 111: Can `invoke` fail for any other thing than non-existent method scenario? Either we can make this a general error, like `HttpStatusCode.INTERNAL_SERVER_ERROR`, or we should only catch errors where a certain string is matched and then return `BAD_REQUEST` for it.

shubham1172 (approved):
Left a comment!

src/implementation/Server/HTTPServer/actor.ts
DeepanshuA: Yup, it can fail for any other reason as well. I have changed it to INTERNAL_SERVER_ERROR.

shubham1172 (approved):
LGTM! We should also do this exercise for other actor methods like deactivate, but it can be done separately, fyi @XavierGeerinck

XavierGeerinck (approved):

XavierGeerinck:
> 

Agreed

Issue #371, opened 2022-09-22
Allow publishing cloud events directly 


shubham1172:
## Describe the proposal

The SDK support publishing `application/json` content-type only. This proposal is to allow sending cloud events directly with content-type as `application/cloudevents+json`.

It is important for users to set CloudEvent attributes for Pub/Sub routing. See #370

See similar implementation by dotnet-SDK https://github.com/dapr/dotnet-sdk/pull/868

PR #428, opened 2022-12-12
Allow publish cloud events directly

shubham1172:

# Description

This change allows publishing cloud events to Dapr. The input data can now be string or object.

1. If the input data is a string, send it as text/plain
2. If the input data is an object and contains the required properties from cloud event spec, send it as a cloud event, else send it as a JSON.

This PR also changes the response from the publish operation. It now returns a custom type with `error` if the publish request failed.

## Issue reference

Please reference the issue this PR will close: #371 

## Checklist

- [ ] Code compiles correctly
- [ ] Created/updated tests


shubham1172:
One downside to this approach is, users can't send a vanilla JSON value which looks like cloud-event, since it's always converted to cloud-event+json type. To avoid this, we can add a explicit method, `publishCloudEvent`. Thoughts? @XavierGeerinck @DeepanshuA

XavierGeerinck:
> One downside to this approach is, users can't send a vanilla JSON value which looks like cloud-event, since it's always converted to cloud-event+json type. To avoid this, we can add a explicit method, `publishCloudEvent`. Thoughts? @XavierGeerinck @DeepanshuA

We only convert to cloud event type when code has `specversion`, `source`, `type` and `id` right? so I don't think this is an issue seeing that these fields are quite "rare" to be used together. Plus this also converts users to the cloud events ecosystem. So personally I am in favor of having this like it's implemented

XavierGeerinck (approved):

Issue #439, opened 2023-01-09
ttlInSeconds metadata not respected when state saved using the client


dgcaron:
## Expected Behavior
When i set the TTL when saving state i expect the state to be removed after the time has expired. This does work using the http api but not with the client. 

## Actual Behavior

when calling the http api 

```
POST http://localhost:3501/v1.0/state/statestore

[{
	"key": "mykey",
	"value": {
		"prop-1": "A",
		"prop-2": "B",
	},
	"metadata": { "ttlInSeconds": "5" }
}]

```
the state is cleared after 5 seconds but when i call the save method 

```
 await this.dapr.state.save(DaprComponents.stateStore, [{
                key: `mykey`,
                value:  data,
                metadata: {
                    ttlInSeconds: "5"
                }
            }]);

```
the state is stored but not cleared.

PR #461, opened 2023-03-10
Allow passing metadata to client.state.save and return a custom response with error

shubham1172:
# Description

- Allow users to pass metadata to client.state.save. 
- State save operation returns a response (that can contain error) instead of void
- Migrates e2e tests from HTTP and gRPC to common e2e tests

Note, users can pass individual metadata per state item, as well as request level metadata. In HTTP, Dapr runtime prioritizes the request metadata over individual metadata. In gRPC, Dapr runtime ignores the request metadata. To be consistent, the SDK's gRPC implementation now also behaves like HTTP, and prioritizes request metadata over individual metadata.

## Issue reference

Please reference the issue this PR will close: #439

## Checklist

- [X] Code compiles correctly
- [x] Created/updated tests


src/implementation/Client/GRPCClient/state.ts
```diff
@@ -45,7 +46,11 @@ export default class GRPCClientState implements IClientState {
     this.client = client;
   }
 
-  async save(storeName: string, stateObjects: KeyValuePairType[]): Promise<void> {
+  async save(
+    storeName: string,
+    stateObjects: KeyValuePairType[],
+    metadata?: KeyValueType,
```
XavierGeerinck, line 52: Shouldn't we make this an option parameter?

src/interfaces/Client/IClientState.ts
```diff
@@ -18,9 +18,10 @@ import { KeyValueType } from "../../types/KeyValue.type";
 import { StateQueryType } from "../../types/state/StateQuery.type";
 import { StateQueryResponseType } from "../../types/state/StateQueryResponse.type";
 import { StateGetBulkOptions } from "../../types/state/StateGetBulkOptions.type";
+import { StateSaveResponseType } from "../../types/state/StateSaveResponseType";
 
 export default interface IClientState {
-  save(storeName: string, stateObjects: KeyValuePairType[]): Promise<void>;
+  save(storeName: string, stateObjects: KeyValuePairType[], metadata?: KeyValueType): Promise<StateSaveResponseType>;
```
XavierGeerinck, line 24: Options param?

XavierGeerinck (changes requested):

src/implementation/Client/GRPCClient/state.ts
shubham1172: We can do that, good idea. Updated.

XavierGeerinck (approved):

Issue #442, opened 2023-01-12
[DaprClient] Refactor all the interfaces to have a options object


XavierGeerinck:
## Describe the proposal

The interfaces are becoming too large in some instances. This is why the proposal is being made to add all required parameters under the interface and move optional parameters to the `options` object

(linked issues: #294) 

Example

```typescript
export default interface IClientInvoker {
  invoke(
    appId: string,
    methodName: string,
    method: HttpMethod,
    data?: object,
    options?: InvokerOptions,
  ): Promise<object>;
}
```

Should become

```typescript
export default interface IClientInvoker {
  invoke(
    appId: string,
    methodName: string,
    method: HttpMethod,
    options?: InvokerOptions,
  ): Promise<object>;
}
```

PR #466, opened 2023-03-20
feat(pubsub): Allow setting explicit content-type during publish operations

shubham1172:
# Description

This PR 
- introduces an options parameter to Publish in DaprClient
- adds an option to serializer to override the content-type (and not always infer it)

## Issue reference

Please reference the issue this PR will close: #442

## Checklist

- [ ] Code compiles correctly
- [ ] Created/updated tests


XavierGeerinck:
Ref offline discussion, let's move the Content-Type overriding to the Serializer to abstract it away so users can override the content type everywhere besides just the pubsub system

PR #471, opened 2023-03-24
Fix/query

### LINKED ISSUE #460: State query returns empty string instead of StateQueryResponseType
## Expected Behavior

If I run a query against my state store, where there should be no results, I should get a `StateQueryResponseType` that has and empty array for the results property.

```javascript
let res = await client.state.query("pg-state", {filter: { EQ: { foo: "bar" } } });
```

## Actual Behavior

Instead it returns an empty string.

salmankhan-prs:
# Description

This PR adds a fix for an issue where the state query method returns an empty string instead of a `StateQueryResponseType` when there are no results to return. The fix adds an `if` statement to handle empty responses and return an empty array instead.

## Issue reference

This PR closes #460.

## Checklist

- [x] Code compiles correctly
- [ ] Created/updated tests

### DIFF
--- a/src/implementation/Client/HTTPClient/state.ts
+++ b/src/implementation/Client/HTTPClient/state.ts
@@ -104,7 +104,7 @@ export default class HTTPClientState implements IClientState {
     });
 
     if (result === "") {
-      return [];
+      return { results: [] };
     }
 
     return result as StateQueryResponseType;
--- a/src/implementation/Client/HTTPClient/state.ts
+++ b/src/implementation/Client/HTTPClient/state.ts
@@ -103,6 +103,10 @@ export default class HTTPClientState implements IClientState {
       },
     });
 
+    if (result === "") {
+      return [];
+    }
+
     return result as StateQueryResponseType;
   }
 }
--- a/src/implementation/Client/HTTPClient/state.ts
+++ b/src/implementation/Client/HTTPClient/state.ts
@@ -104,7 +104,7 @@ export default class HTTPClientState implements IClientState {
     });
 
     if (result === "") {
-      return { results: [] };
+      return { results: [] } as StateQueryResponseType;
     }
 
     return result as StateQueryResponseType;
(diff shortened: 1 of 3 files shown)


src/implementation/Client/HTTPClient/state.ts
```diff
@@ -103,6 +103,10 @@ export default class HTTPClientState implements IClientState {
       },
     });
 
+    if (result === "") {
+      return { results: [] };
```
shubham1172, line 107: This fixes HTTP, please add gRPC, along with tests. You can also return it as `StateQueryResponseType`.
```suggestion
      return { results: [] } as StateQueryResponseType;
```

salmankhan-prs:
@shubham1172  

for the gRPC please check the below code 

```javascript

 const resultsList = res.getResultsList();

        if (resultsList.length === 0) {  // add this if statement to handle empty responses
          return resolve({
            results: [],
            token: res.getToken(),
          });
        }

        const resMapped: StateQueryResponseType = {
          results: resultsList.map((i) => ({
            key: i.getKey(),
            data: i.getData(),
            etag: i.getEtag(),
            error: i.getError(),
          })),
          token: res.getToken(),
        };

        return resolve(resMapped);
```

Regarding the test cases should I create a new test case or edit the test cases that already exist?
where should I create the test case in e2e>comman >client.test.ts?
also, How can I run the test case locally .please provide your input on this

shubham1172:
@salmankhan-prs LGTM. Yes, you can create a new test under "state" in https://github.com/dapr/js-sdk/blob/main/test/e2e/common/client.test.ts. To run the tests locally, you can run "npm run test:e2e:common:client". 

You can see https://github.com/dapr/js-sdk/blob/main/scripts/test-init.sh to run any additional components (like mongodb) for common tests.

shubham1172 (approved):

shubham1172:
LGTM, thanks for the PR @salmankhan-prs!

XavierGeerinck (approved):

PR #493, opened 2023-05-26
[WIP] Add support to release dev packages

shubham1172:
# Description

This PR refactors the build pipelines and adds a job to release dev packages (on push to main).
1. Removes unused script ./scripts/publish.sh
2. Adds a new command to package.json `build-ci` to build typescript package without linting or prettifying
3. Split the github actions into three jobs - build, publish, and publish-dev.

## Issue reference

## Checklist

- [ ] Code compiles correctly
- [ ] Created/updated tests


.github/workflows/build.yml
```diff
@@ -29,20 +29,19 @@ on:
 jobs:
   build:
     runs-on: ubuntu-latest
-    environment: production
-    env:
-      NODE_VER: 16.14.0
+    strategy:
+      fail-fast: false
+      matrix:
```
DeepanshuA, line 34: I like having matrix here to be able to build with multiple node versions.
Do we want to add any other node version here?

.github/workflows/build.yml
```diff
@@ -53,30 +52,87 @@ jobs:
...
+        env:
+          NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
+
+  publish-dev:
+    needs: build
+    if: startsWith(github.ref, 'refs/heads/main')
+    runs-on: ubuntu-latest
+    environment: development
+    env:
+      NODE_VER: 16.14.0
```
DeepanshuA, line 105: Can we take input of node version from a single place in build.yml, instead of providing it at multiple places, like build, publish, publish-dev ?

.github/workflows/build.yml
```diff
@@ -53,30 +52,87 @@ jobs:
...
-      - name: Is Release?
-        if: startswith(github.ref, 'refs/tags/v')
-        run: echo "DEPLOY_PACKAGE=true" >> $GITHUB_ENV
+  publish:
+    needs: build
+    if: startswith(github.ref, 'refs/tags/v')
+    runs-on: ubuntu-latest
+    environment: production
+    env:
+      NODE_VER: 16.14.0
```
DeepanshuA, line 61: So, we can build with multiple versions, but will publish with only one version?

scripts/build.sh
```diff
@@ -21,9 +21,6 @@ echo "Executing in $(pwd)"
 echo "Description: Build the package in build/"
 echo "====================================================="
 
-echo "Installing Dependencies"
-npm install > /dev/null
```
DeepanshuA, line 25: Should we let `npm install` be a part of build.sh only, as we need to do `npm install` separately everywhere?

.github/workflows/build.yml
```diff
@@ -53,30 +52,87 @@ jobs:
...
 
-      # note: package.json gets updated here for the new package name
       - name: "[dapr-client] Publish to npm (dapr-client)"
-        if: env.DEPLOY_PACKAGE == 'true'
+        run: npm publish build/ --access public
+        env:
+          NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
+
+  publish-dev:
+    needs: build
```
DeepanshuA, line 100: Please help me understand: `publish-dev` needs `build` job to be done, and in `build` job, we have `Build package (install dependencies, lint, prettier, build)` - so, are we checking lint and prettier for publish-dev as well?

.github/workflows/build.yml
shubham1172: Yes Deepanshu, I agree, added this to track https://github.com/dapr/js-sdk/issues/494

.github/workflows/build.yml
shubham1172: Build is for testing (and running UTs), but publish can work with any node version. We don't have `engines` specified in our package.json as of now https://docs.npmjs.com/cli/v9/configuring-npm/package-json?v=true#engines, so as long as the dependencies work, the published package should be able to run with any node. Added a comment to #494 to explore adding this to our package.json

.github/workflows/build.yml
shubham1172: Shared b/w publish and publish-dev, skipped for build due to https://github.com/orgs/community/discussions/26388#discussioncomment-3251731

scripts/build.sh
shubham1172: The problem is that running lint/prettier also requires `npm install`. So if we added this step inside `build.sh`, you will only be able to lint AFTER the build, but we want to fail fast. In package.json, we are doing this now

````
"build": "npm install && npm run lint && npm run pretty && npm run build-ci",
```

.github/workflows/build.yml
shubham1172: Added the `needs` dependency to run it sequentially, so that if build (lint/pretty) or unit tests fail, we don't even start publishing.

DeepanshuA (approved):

XavierGeerinck (approved):
Ok for me! Question: don't we want to use the commit sha hash instead of a timestamp?

shubham1172:
> Ok for me! Question: don't we want to use the commit sha hash instead of a timestamp?

Good idea, I think SHA will be helpful when users want to associate dev release with a commit - I had kept timestamps only for one reason, easy to find the latest version and based on time. Let me add a go style versioning in that case.

e.g. 3.0.1-20230526200301-0744d001aa84

WDYT @XavierGeerinck?

XavierGeerinck:
I would personally just do the sha hash :)

shubham1172:
Since short hashes can have collisions, we should either use
1. timestamp + short hash
2. long hash

XavierGeerinck:
> Since short hashes can have collisions, we should either use
> 
> 1. timestamp + short hash
> 2. long hash

Good point, ok then we can use your proposal

XavierGeerinck (approved):

Issue #537, opened 2023-10-18
ActorId accepts invalid IDs, causing issues when making Actor HTTP calls


heunghingwan:
When constructing an actor using strings such as url, it will throw with app-id not found error. I don’t know whether it is an SDK or core error.

## Steps to Reproduce the Problem

`const actor = await builder.build(new ActorId(url))`
`actor.test()` <- throw with app-id not found

shubham1172:
https://github.com/dapr/js-sdk/blob/b063e10e4b1a122fe7bc266d1681ca0727b6a522/src/actors/client/ActorClient/ActorClientHTTP.ts#L50

shubham1172:
So URLs should not be allowed as actor IDs because of the way actor APIs are written, reference: https://docs.dapr.io/reference/api/actors_api/#get-actor-reminder

For example, if actor ID is https://foo.bar, the below request will look like `/actors/${actorType}/https://foo.bar/state/${key}`
https://github.com/dapr/js-sdk/blob/b063e10e4b1a122fe7bc266d1681ca0727b6a522/src/actors/client/ActorClient/ActorClientHTTP.ts#L50

What we should do is throw an error if the actor ID is not proper or potentially sanitize it, instead of simply accepting any string.
https://github.com/dapr/js-sdk/blob/b063e10e4b1a122fe7bc266d1681ca0727b6a522/src/actors/ActorId.ts#L20

XavierGeerinck:
> Does this mean it will work when using grpc protocol?

It would, but gRPC is not supported on the runtime :)

PR #539, opened 2023-10-18
Add Actor ID validation

heunghingwan:
# Description

Throw when actor id contain '/'

## Issue reference

issue this PR will close: #537 

## Checklist

- [x] Code compiles correctly
- [x] Created/updated tests

### DIFF
--- a/src/actors/ActorId.ts
+++ b/src/actors/ActorId.ts
@@ -17,6 +17,7 @@ export default class ActorId {
   private readonly id: string;
 
   constructor(id: string) {
+    if (id.match(/\//)) throw new Error("ActorId cannot contain '/'");
     this.id = id;
   }
 
--- a/src/actors/ActorId.ts
+++ b/src/actors/ActorId.ts
@@ -18,7 +18,7 @@ export default class ActorId {
 
   constructor(id: string) {
     if (!id) {
-        throw new Error("ActorId cannot be empty");
+      throw new Error("ActorId cannot be empty");
     }
     this.id = id;
   }
--- a/src/actors/ActorId.ts
+++ b/src/actors/ActorId.ts
@@ -17,7 +17,9 @@ export default class ActorId {
   private readonly id: string;
 
   constructor(id: string) {
-    if (!id) throw new Error("ActorId cannot be empty");
+    if (!id) {
+        throw new Error("ActorId cannot be empty");
+    }
     this.id = encodeURIComponent(id);
   }
 
--- a/src/actors/ActorId.ts
+++ b/src/actors/ActorId.ts
@@ -17,6 +17,7 @@ export default class ActorId {
   private readonly id: string;
 
   constructor(id: string) {
+    if(id==='') throw new Error("ActorId cannot be empty")
     if (id.match(/\//)) throw new Error("ActorId cannot contain '/'");
     this.id = id;
   }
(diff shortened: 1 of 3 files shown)


heunghingwan:
Add empty checking after some reading in dapr/dapr#6461

> Take for example the route v1.0/state/{storeName}/{key}. Before, passing an empty {storeName}, such as v1.0/state//somekey would have still invoked the handler, which would have received an empty "storeName" (and likely returned 400 - bad request). Because the router now normalizes paths and removes double slashes (see https://github.com/dapr/dapr/issues/6324), the path is converted to v1.0/state/somekey, so it now returns a 404 because it can't match any route.

src/actors/ActorId.ts
```diff
@@ -17,6 +17,8 @@ export default class ActorId {
   private readonly id: string;
 
   constructor(id: string) {
+    if(id==='') throw new Error("ActorId cannot be empty")
```
XavierGeerinck, line 20: I would say this is not generic enough. In Javascript we should use `encodeURIComponent` to ensure the strings are valid. Maybe it would be good to use something like this:

```javascript
if (!id) {
    throw new Error("ActorId cannot be empty");
}

id = encodeURIComponent(id);
```

XavierGeerinck (changes requested):
Thank you for adding a test!!! Almost there, just a small change :)

src/actors/ActorId.ts
heunghingwan: Will this cause compatibility issues with other language SDKs?

XavierGeerinck:
Can we still change `if (id === "") throw new Error("ActorId cannot be empty");` ?

heunghingwan:
> Can we still change `if (id === "") throw new Error("ActorId cannot be empty");` ?

Yes, it will be more versatile when importing in a javascript-only application

XavierGeerinck:
> > Can we still change `if (id === "") throw new Error("ActorId cannot be empty");` ?
> 
> Yes, it will be more versatile when importing in a javascript-only application

Perfect! But please use the multi line if instead of single one

```javascript
if (!id) {
...
}
```

shubham1172:
@heunghingwan were you able to test your scenario from the issue with this PR? 

https://github.com/dapr/js-sdk/blob/b063e10e4b1a122fe7bc266d1681ca0727b6a522/src/actors/client/ActorClient/ActorClientHTTP.ts#L50

Since we are decoding back in `getId`, it should still not work since the ID will contain special characters again.

heunghingwan:
> @heunghingwan were you able to test your scenario from the issue with this PR?
> 
> https://github.com/dapr/js-sdk/blob/b063e10e4b1a122fe7bc266d1681ca0727b6a522/src/actors/client/ActorClient/ActorClientHTTP.ts#L50
> 
> Since we are decoding back in `getId`, it should still not work since the ID will contain special characters again.

You are right, maybe add a getURISafeId function, or encoded in ActorClientHTTP.ts, to provide unencoded id for grpc use?

XavierGeerinck (approved):

shubham1172:
@XavierGeerinck one issue I see with this change is, we are introducing a parity between HTTP and gRPC actor clients. HTTP will use a sanitized actor ID, and gRPC won't. This means there will be interop issues b/w the two protocols. 

I would propose to modify gRPC's behavior too and keep it consistent. Thoughts?

shubham1172:
@heunghingwan I see, it should be fine in that case, thanks for your contribution!

Issue #547, opened 2023-10-31
Accept the api token through the `DAPR_API_TOKEN` environment variable


elena-kolevska:
As described in https://github.com/dapr/proposals/blob/main/0010-S-unified-api-token-env-variable.md

PR #548, opened 2023-10-31
If an api token is not passed in the constructor, check the env variable

elena-kolevska:
# Description

Adds the possibility to get an api token from the standard environment variable, as described in this proposal https://github.com/dapr/proposals/blob/main/0010-S-unified-api-token-env-variable.md

## Issue reference
https://github.com/dapr/js-sdk/issues/547

## Checklist

- [X] Code compiles correctly
- [X] Created/updated tests
- [X] Extended the documentation

### DIFF
--- a/src/utils/Client.util.ts
+++ b/src/utils/Client.util.ts
@@ -292,7 +292,7 @@ export function getClientOptions(
     isKeepAlive: clientOptions?.isKeepAlive,
     logger: clientOptions?.logger ?? defaultLoggerOptions,
     actor: clientOptions?.actor,
-    daprApiToken: clientOptions?.daprApiToken,
+    daprApiToken: clientOptions?.daprApiToken ?? process.env.DAPR_API_TOKEN,
     maxBodySizeMb: clientOptions?.maxBodySizeMb,
   };
 }
--- a/src/utils/Client.util.ts
+++ b/src/utils/Client.util.ts
@@ -292,7 +292,7 @@ export function getClientOptions(
     isKeepAlive: clientOptions?.isKeepAlive,
     logger: clientOptions?.logger ?? defaultLoggerOptions,
     actor: clientOptions?.actor,
-    daprApiToken: clientOptions?.daprApiToken ?? process.env.DAPR_API_TOKEN,
+    daprApiToken: clientOptions?.daprApiToken ?? Settings.getDefaultApiToken(),
     maxBodySizeMb: clientOptions?.maxBodySizeMb,
   };
 }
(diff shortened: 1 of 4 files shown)


src/utils/Client.util.ts
```diff
@@ -292,7 +292,7 @@ export function getClientOptions(
     isKeepAlive: clientOptions?.isKeepAlive,
     logger: clientOptions?.logger ?? defaultLoggerOptions,
     actor: clientOptions?.actor,
-    daprApiToken: clientOptions?.daprApiToken,
+    daprApiToken: clientOptions?.daprApiToken ?? process.env.DAPR_API_TOKEN,
```
shubham1172, line 295: We can move this to [Settings.util.ts](https://github.com/dapr/js-sdk/blob/0fb83c1fda6a47e3f9a5753d24e6b0dcfe311985/src/utils/Settings.util.ts#L93) for consistency.

test/unit/utils/Client.util.test.ts
```diff
@@ -359,6 +359,36 @@ describe("Client.util", () => {
...
+
+    it("returns correct Dapr Client Options when token provided in constructor", () => {
+      const inOptions: Partial<DaprClientOptions> = {
+        daprApiToken: "token",
+      };
+      const options = getClientOptions(inOptions, CommunicationProtocolEnum.HTTP, undefined);
+      expect(options.daprApiToken).toEqual("token");
+    });
+
+    it("returns correct Dapr Client Options when token provided in anv variable", () => {
```
shubham1172, line 371:
```suggestion
    it("returns correct Dapr Client Options when token provided in env variable", () => {
```

test/unit/utils/Client.util.test.ts
```diff
@@ -359,6 +359,36 @@ describe("Client.util", () => {
...
+      const oldToken = process.env.DAPR_API_TOKEN
+      process.env.DAPR_API_TOKEN = "envtoken";
+      const options = getClientOptions(undefined, CommunicationProtocolEnum.HTTP, undefined);
+      expect(options.daprApiToken).toEqual("envtoken");
+      process.env.DAPR_API_TOKEN = oldToken;
+    });
+
+    it("returns correct Dapr Client Options when token provided both in constructor and in anv variable", () => {
```
shubham1172, line 379:
```suggestion
    it("returns correct Dapr Client Options when token provided both in constructor and in env variable", () => {
```

shubham1172 (changes requested):
Thanks for the PR @elena-kolevska. Small nits and the lint is failing.

src/utils/Client.util.ts
elena-kolevska: I thought about the same thing, but it seemed like we only use it when there is a default value. Let me try it out and see how it looks.

shubham1172: Yeah, I noticed that in some cases we even have `""` as the default, which essentially is the same as not having one.

shubham1172 (approved):

PR #553, opened 2023-11-20
fix flaky integration test DemoActorReminderOnceImpl

MregXN:
# Description
it fails due to the following log
```
 test/actor/DemoActorReminderOnceImpl.ts:24:7 - error TS2345: Argument of type 'undefined' is not assignable to parameter of type 'Duration'.
```

the `period` should be able to be undefined.

## Issue reference

Please refer to the the point 4 of issue [#7128](https://github.com/dapr/dapr/issues/7128) in dapr/dapr.


shubham1172:
Thank you so much for looking into this @MregXN, these improvements will go a long way in increasing the SDK e2e test stability.

src/actors/runtime/AbstractActor.ts
```diff
@@ -95,8 +95,8 @@ export default abstract class AbstractActor {
   async registerActorReminder<_Type>(
     reminderName: string,
     dueTime: Temporal.Duration,
-    period: Temporal.Duration,
-    ttl?: Temporal.Duration,
+    period: Temporal.Duration | undefined,
+    ttl?: Temporal.Duration | undefined,
```
shubham1172, line 99: `ttl?` makes the type optional, so ` | undefined` is not required https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#optional-properties

We can also make `period` as an optional type.

[playground link](https://www.typescriptlang.org/play?#code/GYVwdgxgLglg9mABAcwE4FN1QBRgIYC26AXIgM5SoxjKIA+i4AJusNekwJSIDeAUIkGIICMnAA26AHTi4ybAAMAEunGyANIgAkPfEQC+CzgG4++vn1CRYCFBiwB1GFAAWAeQAON-OIAKeVEJcQnQAflIKKhpufiFhUQlpWXllVQ1tXRDDEzMLNEwcACIAKTgXMEKc-KxsZlZ2LlM+aqgnV09vPD8AoJKyiqr7Vud3L3gff0CCWrAWNjAOQYK20c7uqewTIA)

MregXN: Thanks, you are right

MregXN:
ready to review

PR #555, opened 2023-11-28
fix e2e test common/server/server.test.ts

MregXN:
# Description
It fails due to the following log. For more log details refer to this [link](https://github.com/dapr/dapr/actions/runs/7017990038/job/19092429905).
```
...
  ● common/server › pubsub › grpc/should mark messages as processed successfully (SUCCESS) by-default, and the same message should not be received anymore

    expect(received).toBe(expected) // Object.is equality

    Expected: 1
    Received: 0

      163 |         // Delay a bit for event to arrive
      164 |         await new Promise((resolve, _reject) => setTimeout(resolve, 1000));
    > 165 |         expect(mockSubscribeStatusHandler.mock.calls.length).toBe(1);
          |                                                              ^
      166 |         expect(mockSubscribeStatusHandler.mock.calls[0][1]).toEqual("SUCCESS");
      167 |         expect(mockSubscribeDeadletterHandler.mock.calls.length).toBe(0);
      168 |       },

      at test/e2e/common/server.test.ts:165:62
      at fulfilled (test/e2e/common/server.test.ts:17:58)
...
```

The delay is not long enough to wait for the event to arrive. Adding more delay could be helpful. 

## Issue reference

Please refer to the the point 4 of issue [#7128](https://github.com/dapr/dapr/issues/7128) in dapr/dapr.

### DIFF
--- a/test/e2e/common/server.test.ts
+++ b/test/e2e/common/server.test.ts
@@ -138,7 +138,8 @@ describe("common/server", () => {
     });
 
     await setupPubSubSubscriptions();
-    // Sleep for 2 seconds to allow all subscriptions to be registered
+    // Sleep to make the tests less flaky.
+    // TODO: https://github.com/dapr/js-sdk/issues/560
     await NodeJSUtil.sleep(2000);
 
     await httpServer.start();
--- a/test/e2e/common/server.test.ts
+++ b/test/e2e/common/server.test.ts
@@ -12,6 +12,7 @@ limitations under the License.
 */
 
 import { CommunicationProtocolEnum, DaprServer, DaprPubSubStatusEnum } from "../../../src";
+import * as NodeJSUtil from "../../../src/utils/NodeJS.util";
 
 const serverHost = "127.0.0.1";
 const serverGrpcPort = "50001";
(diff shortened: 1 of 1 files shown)


test/e2e/common/server.test.ts
```diff
@@ -134,9 +138,13 @@ describe("common/server", () => {
     });
 
     await setupPubSubSubscriptions();
+    // Sleep for 2 seconds to allow all subscriptions to be registered
+    await NodeJSUtil.sleep(2000);
```
shubham1172, line 142: Is this really required? When we `register` a subscription, it just gets added to an internal data structure. Only after the server starts below, and Dapr sidecar calls `/dapr/subscribe` on app port, we return the list of subscriptions to the runtime which registers the subscriptions actually.

shubham1172: That is interesting why this works, do you mind running this locally and see logs (with timestamps) on both gRPC and HTTP based sidecars? I think that should give us more idea.

test/e2e/common/server.test.ts
```diff
@@ -134,9 +138,13 @@ describe("common/server", () => {
     });
 
     await setupPubSubSubscriptions();
+    // Sleep for 2 seconds to allow all subscriptions to be registered
```
shubham1172, line 141:
```suggestion
    // Sleep to make the tests less flaky.
    // TODO: https://github.com/dapr/js-sdk/issues/560
```

test/e2e/common/server.test.ts
shubham1172: Resolved this with the comment above.

shubham1172 (approved):

shubham1172:
Closing this PR to fix flakiness in tests for now, root cause still remains to be inspected and will be tracked by #560

Issue #570, opened 2024-02-08
Fetching Keys from Azure App Config using metadata.label not returning appropriate response.


dbauswel:
Fetching Keys from Azure App Configuration using metadata.label field does not work. Currently receiving a 404 response.
code snippet attached.

```
const config = await client.configuration.get(
            'azureappconfig',
            [key1],
            {
                metadata: {
                    label: yourEnv
                }
            }
        )
```

Note that making a direct call works.
`/v1.0/configuration/azureappconfig?key=key1&metadata.label=yourEnv`

shubham1172:
Hi @dbauswel, the HTTP DaprClient [does not support configuration](https://github.com/dapr/js-sdk/blob/main/src/implementation/Client/HTTPClient/configuration.ts) so I guess you meant the gRPC DaprClient is throwing a not-found error? Are you able to make a direct gRPC request and get this data?

tlund101:
> Hi @dbauswel, the HTTP DaprClient [does not support configuration](https://github.com/dapr/js-sdk/blob/main/src/implementation/Client/HTTPClient/configuration.ts) so I guess you meant the gRPC DaprClient is throwing a not-found error? Are you able to make a direct gRPC request and get this data?

@shubham1172 I'm familiar with the issue, and this is correct.  Both GetConfiguration and GetConfigurationAlpha1 from the proto file correctly return a key with the matching metadata label via gRPC request.  Trying to use the gRPC DaprClient seems to ignore the metadata and returns a matching key without a label (if there is one, otherwise it logs a 404 error).

Passing an empty array to fetch all keys returns them all, and those that have labels show them in the metadata field that is returned.

PR #571, opened 2024-02-09
Fix Configuration GRPC calls not sending metadata

tlund101:
# Description

When sending a request for Configurations from Azure App Config, the call would silently drop the metadata that was sent with it.  This would cause keys to be returned only if they had no labels on them when specified.

Having some unrelated errors preventing my local tests from running, though I did validate the change with the app that created this issue.  Let me know if I'm missing a spot that should also be updated, as I'm new to this codebase.

## Issue reference

Please reference the issue this PR will close: #570 

## Checklist

- [x] Code compiles correctly
- [x] Created/updated tests

### DIFF
--- a/src/implementation/Client/GRPCClient/configuration.ts
+++ b/src/implementation/Client/GRPCClient/configuration.ts
@@ -12,7 +12,6 @@ limitations under the License.
 */
 
 import GRPCClient from "./GRPCClient";
-import * as grpc from "@grpc/grpc-js";
 import {
   GetConfigurationRequest,
   GetConfigurationResponse,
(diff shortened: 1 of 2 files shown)


src/implementation/Client/GRPCClient/configuration.ts
```diff
@@ -51,6 +51,7 @@ export default class GRPCClientConfiguration implements IClientConfiguration {
       for (const [key, value] of Object.entries(metadataObj)) {
         metadata.add(key, value);
```
shubham1172, line 52: We do not need to pass `metadata` anymore to the call since we are populating `msg.getMetadataMap()`.

shubham1172, line 52: We do not need to pass `metadata` anymore to the call since we are populating `msg.getMetadataMap()`, so we can cleanup the `metadata` variable (at both places).

shubham1172 (changes requested):
Thanks @tlund101 for the fast PR! 🚀 Just a small comment.

src/implementation/Client/GRPCClient/configuration.ts
tlund101: Thanks, I wasn't sure if the parameter was used for something that the message metadata wasn't used for.  Fixed!

shubham1172 (approved):
LGTM, let's #ShipIt 🚀

Issue #584, opened 2024-03-08
Next.js DaprServer Module not found: Can't resolve 'crypto'


majian159:
## Actual Behavior

I'm attempting to start the DaprServer in instrumentation.ts in a Next.js project, but it's throwing an exception:
```console
○ Compiling /instrumentation ...
 ⨯ ./node_modules/.pnpm/@dapr+dapr@3.3.1/node_modules/@dapr/dapr/actors/ActorId.js:15:0
Module not found: Can't resolve 'crypto'

https://nextjs.org/docs/messages/module-not-found

Import trace for requested module:
./node_modules/.pnpm/@dapr+dapr@3.3.1/node_modules/@dapr/dapr/index.js
./src/instrumentation.ts
```
The code is as follows:
```typescript
import { CommunicationProtocolEnum, DaprServer } from "@dapr/dapr";

export async function register() {
  const server = new DaprServer({
    serverPort: "4000",
    communicationProtocol: CommunicationProtocolEnum.GRPC,
  });

  await server.start();
}
```

shubham1172:
![image](https://github.com/dapr/js-sdk/assets/22132944/685919be-4b8e-44df-a340-5ea225f41375)

Does this help?

https://nextjs.org/docs/messages/module-not-found#the-module-youre-trying-to-import-uses-nodejs-specific-modules

majian159:
This is not helpful; I am using the instrumentation.ts method with Next.js 14 to start the Dapr server.

This is a minimal reproduction repository: https://github.com/majian159/dapr-server-example

shubham1172:
Thanks for sharing this @majian159, I am out this week but I/someone will check this soon.

xAmast:
@shubham1172 My latest pull request replaces the crypto module with [crypto](https://developer.mozilla.org/zh-CN/docs/Web/API/crypto_property). This change will benefit all runtimes compliant with the WinterCG specification (including Next.js [Edge Runtime](https://nextjs.org/docs/app/api-reference/edge#crypto-apis)).

Issue #494, opened 2023-05-26
Document and test select node versions


shubham1172:
## Describe the proposal

We do not have officially supported versions today and our tests always run on a single node version (16.x.x for now).

<img width="453" alt="image" src="https://github.com/dapr/js-sdk/assets/22132944/982820e6-56e1-46f2-b3b4-95c1db60ad0a">

With this proposal, we should maintain a list of node versions that we officially support and test them.

shubham1172:
Also explore if an engines entry should be added - https://docs.npmjs.com/cli/v9/configuring-npm/package-json?v=true#engines

shubham1172:
Node 16 has reached EOL on Sep 2023, we should move to Node 18+ https://nodejs.org/en/blog/announcements/nodejs16-eol

PR #593, opened 2024-05-01
feat: Replace crypto module with crypto object associated with the global  object.

xAmast:
# Description

Upgraded TypeScript / ESLint / Prettier.
Updated @types/node to version 18.
Replaced the crypto module with [crypto](https://developer.mozilla.org/zh-CN/docs/Web/API/crypto_property).

## Issue reference

Please reference the issue this PR will close: #584 

## Checklist

- [x] Code compiles correctly
- [ ] Created/updated tests


xAmast:
At present, the Dapr JavaScript SDK only supports the Express framework for Node.js. I intend to expose the `server.handle: (req: Request) => Promise<Response>` interface in the next PR to support other frameworks and runtimes (ensuring that Request/Response adhere to the WinterCG specification).

## Integration example

**Bun integration**
``` javascript
import { DaprServer } from "@dapr/dapr";

const daprServer = new DaprServer();

const server = Bun.serve({
  port: 3000,
  fetch(req) {
    return daprServer.handle(req);
  },
});
```

**Deno integration**
``` javascript
import { DaprServer } from "npm:@dapr/dapr";

const daprServer = new DaprServer();

Deno.serve(async (_req: Request) => {
    const res = await daprServer.handle(req);
    return res;
});
```

**Next.js integration**

Create app/[[...slugs]]/route.ts inside app router

``` javascript
// app/[[...slugs]]/route.ts
import { DaprServer } from "@dapr/dapr";

const daprServer = new DaprServer();

export const GET = daprServer.handle;
export const POST = daprServer.handle;
```

**Astro integration**

Create pages/[...slugs].ts

``` javascript
import { DaprServer } from "@dapr/dapr";

const daprServer = new DaprServer();

const handle = ({ request }: { request: Request }) => daprServer.handle(request);

export const GET = daprServer.handle;
export const POST = daprServer.handle;
```

shubham1172:
First of all, thanks for this very important PR and your goal to make this library Winter CG compatible. In this PR, we are replacing "crypto" the library, with a global import. However, I see from the official docs that this is only available (as stable) from Node 20: https://nodejs.org/docs/latest-v20.x/api/webcrypto.html#web-crypto-api. Dapr JS SDK supports Node 16+ (16 is EOL since a few months now, and we should move to 18+), but the crypto import still does not seem to be stable for node 18.

We can 
- Use #494 to track node upgrade first to 18
- Explore using an alternate library like https://github.com/uuidjs/uuid#readme, instead of using the experimental crypto lib in node 18

/cc @XavierGeerinck

xAmast:
@shubham1172  I'm very sorry, I've been using Bun and Node.JS 20 all along, thus overlooking this point. I'll resubmit the code later.

Issue #417, opened 2022-10-27
Actor client SDK requires actor implementation


danielgerlag:
It seems the to use the JS SDK for actors, the client invoking a remote actor needs a copy of the implementation code in order to register the actor... drilling into the code, it looks like the only reason for this is to get the name of the actor type.  Would it be possible to have an overload to create the proxy builder that does not require passing in the implementation, but rather just the actor type name?

jasonvasquez:
Just to add, this detail is important to me as well.  Within a larger application suite, I am attempting to expose an actor-enabled library to the rest of the application without exposing the implementation.  As a work-around, I now have created client-side "fake" Actor implementations that share the implementation name and implement the Actor's interface.  This works, but isn't ideal.

If, as an alternative, we could pass a constant identifying the Actor, that would work fine here.

PR #697, opened 2025-05-01
chore: add possibility to create actors without the implementation

XavierGeerinck:
# Description

Long overdue, this PR solves the issue that when you are creating an Actor Proxy, that the implementation of the actor was required. There was a work around for it, but this PR makes it more concrete. 

This PR also adds the possibility to use an interface class name that differs from the registered actor.

## Issue reference

Please reference the issue this PR will close: #417

## Checklist

- [x] Code compiles correctly
- [x] Created/updated tests


WhitWaldo:
At first glance, this appears to implement a breaking change as anyone currently using actors would have to rebuild their applications to support this. Is that intentional?

XavierGeerinck:
Added backwards-compatibility for the breaking change :)

PR #714, opened 2025-08-19
Fixed: use grpc-tools protoc to restore --js_out on modern protoc

ConstantinChirila:
# Description

### Problem:
Running `scripts/fetch-proto.sh` failed with: `--js_out: protoc-gen-js: program not found or is not executable`
On modern protoc versions, the builtin JS generator is no longer available the same way, so `--js_out` tries to invoke an external `protoc-gen-js` that isn’t in PATH.

This is due protoc v28 no longer providing the legacy/bundled JS generator, so `--js_out` requires a plugin discovery step that fails in current environment. `grpc-tools` ships a `protoc` binary and plugins known to work with Node’s JS/TS generation this Eliminates PATH/plugin discovery issues.

### The change:
- Prefer `node_modules/.bin/grpc_tools_node_protoc` for generation; fall back to `protoc` if unavailable.
- Keep explicit plugin paths for TS and gRPC: `--plugin=protoc-gen-ts=.../protoc-gen-ts` and `--plugin=protoc-gen-grpc=.../grpc_tools_node_protoc_plugin`

## Issue reference

No issue reference. Needed to unblock further work.

## Checklist

- [ ] Code compiles correctly
- [ ] Created/updated tests


scripts/fetch-proto.sh
```diff
@@ -131,6 +136,8 @@ downloadFile "https://raw.githubusercontent.com/$ORG_NAME/$REPO_NAME/$BRANCH_NAM
 downloadFile "https://raw.githubusercontent.com/$ORG_NAME/$REPO_NAME/$BRANCH_NAME/dapr/proto/internals/v1/apiversion.proto" "$PATH_ROOT/src/proto/dapr/proto/internals/v1/apiversion.proto"
```
WhitWaldo, line 136: We can remove all of these except those actually used by the SDK:
dapr.proto
common.proto
appcallback.proto

scripts/fetch-proto.sh
```diff
@@ -147,6 +154,8 @@ echo ""
 echo "Compiling gRPC files"
 generateGrpc "$PATH_ROOT/src/proto" "dapr/proto/common/v1/common.proto"
 generateGrpc "$PATH_ROOT/src/proto" "dapr/proto/internals/v1/apiversion.proto"
+# Also generate code for reminders to satisfy imports
+generateGrpc "$PATH_ROOT/src/proto" "dapr/proto/internals/v1/reminders.proto"
 generateGrpc "$PATH_ROOT/src/proto" "dapr/proto/internals/v1/service_invocation.proto"
```
WhitWaldo, line 159: Same here - we shouldn't actually be using any of these except appcallback.proto, dapr.proto and common.proto as they're internal to the runtime and subject to change at any point.

WhitWaldo (commented):
Left some notes about some protos to remove down the road, but it sounds from the comments that there are inappropriate imports coming from them.

Generally, only appcallback.proto, dapr.proto and common.proto should be referenced by the SDK. All the others are protos specific to the Dapr runtime internally and subject to change at any point. They should not be relied upon for type support.

WhitWaldo:
This looks great @ConstantinChirila - ping me when you've got the tests passing and I can merge this and your other PR and get an RC out the door for this SDK.

WhitWaldo:
@ConstantinChirila I'm going to close this PR as it's no longer relevant. The SDK now uses Buf because I experienced similar issues getting the prototype types to generate properly on Windows. Thank you for taking the time to put this together though and I hope you'll consider future contributions as well!

PR #715, opened 2025-08-20
Attempt to replace generated proto files with Buf

tscolari:
# Description

This is a poc of how to use buf to generate the proto files from the remote repo.
It's similar to the experimentation from the go-sdk.

In a nutshell:
1. Using buf, the protos for generation are referenced remotely from dapr/dapr - so not copied locally anymore
2. There's a manual "lock file" that ensures that the generation is happening against a committed remote sha (this could be replaced by buf's own locking mechanism if dapr/dapr supports it).
3. Ideally CI would ensure that the generated files are up to date and unmodified.

JS is not a language I'm familiar with, but I tried to keep the generation using the same pattern (ts/es), but different outputs can be configured by using/combining different plugins in buf.gen.yaml.

This is not currently generating proto dependencies like google timestamppb for example - I think there's probably a way to generate them, but at the same time I think they can be consumed versioned from https://www.npmjs.com/package/google-protobuf

## Issue reference

## Checklist

- [ ] Code compiles correctly
- [ ] Created/updated tests


WhitWaldo:
@tscolari Thank you very much for putting this together.

I'm working on an overhaul now that generates the protos types using Buf based on your example and I'll cite you in the description when I create it, but.. 

1) I was able to get all the adjacent and imported types to build correctly through and through
2) Because Buf produces a different API shape than protoc did, it's going to require overhauling all the gRPC logic to use the new approach.

As such, I'm not going to be able to merge in your changes, but I certainly do appreciate you taking the time to share this new approach with me as it's definitely similar to the route we'll take going forward here.

PR #745, opened 2026-04-10
Support building on windows + migration to Buf for prototype type generation

WhitWaldo:
# Description

Updating the build scripts to use JavaScript instead of bash scripts so they're compatible on Windows (without resorting to use WSL). Further, because of issues using gRPC proto generation on Windows, I'm migrating to use Buf instead - updated all the types and tests to reflect this shift. No API changes - should be an entirely internal shift.

Supersedes #737 

## Issue reference

## Checklist

- [X] Code compiles correctly
- [X] Created/updated tests


WhitWaldo:
Closing in favor of #752

PR #856, opened 2026-09-17
feat(pubsub): promote BulkPublishEventAlpha1 to stable BulkPublishEvent

### LINKED ISSUE #786: feat(pubsub): Promote BulkPublishEventAlpha1 to stable BulkPublishEvent
## Summary

The Dapr runtime has promoted the `BulkPublishEventAlpha1` RPC to the stable `BulkPublishEvent` endpoint. The JS SDK still targets the alpha1 variant in both its gRPC and HTTP implementations. We need to migrate to the stable API while maintaining backward compatibility during a deprecation period.

## Current State

The SDK currently uses the alpha1 endpoints:
- **gRPC client** (`src/implementation/Client/GRPCClient/pubsub.ts`): calls `client.bulkPublishEventAlpha1(...)` via ConnectRPC
- **HTTP client** (`src/implementation/Client/HTTPClient/pubsub.ts`): hits the `v1.0-alpha1/publish/bulk/{pubSubName}/{topic}` endpoint
- **Proto definition** (`src/proto/dapr/proto/runtime/v1/dapr.proto`): defines both `BulkPublishEventAlpha1` (alpha) and `BulkPublishEvent` (stable)
- **ConnectRPC stubs** (`src/proto/dapr/proto/runtime/v1/dapr_connect.js`): only exposes `bulkPublishEventAlpha1`

The stable `BulkPublishEvent` RPC is already defined in the proto but is not yet wired up in the ConnectRPC service definition or called from the SDK.

## Proposed Changes

1. **Regenerate / update ConnectRPC stubs** to include the stable `bulkPublishEvent` method in the `Dapr` service definition.
2. **Update gRPC client** (`src/implementation/Client/GRPCClient/pubsub.ts`):
   - Change `client.bulkPublishEventAlpha1(...)` → `client.bulkPublishEvent(...)`
3. **Update HTTP client** (`src/implementation/Client/HTTPClient/pubsub.ts`):
   - Change API version from `v1.0-alpha1` to `v1.0`
   - Keep path as `/publish/bulk/{pubSubName}/{topic}`
4. **Public API surface** (`IClientPubSub.publishBulk()`): no signature change needed — the method name and types remain the same for consumers.
5. **Deprecation notice**: If we want to support older Dapr runtimes that only have the alpha1 endpoint, consider a fallback or a minimum-Dapr-version note in the docs.
6. **Update tests**: Ensure E2E tests exercise the stable endpoint (requires Dapr runtime ≥ the version that promoted BulkPublishEvent to stable).

## Acceptance Criteria

- [ ] `client.pubsub.publishBulk(...)` calls the stable `BulkPublishEvent` gRPC RPC
- [ ] HTTP client uses `v1.0/publish/bulk/...` instead of `v1.0-alpha1/publish/bulk/...`
- [ ] ConnectRPC generated stubs include `bulkPublishEvent`
- [ ] Existing unit and E2E tests pass
- [ ] Documentation updated to reflect stable API usage
- [ ] CHANGELOG entry noting the promotion from alpha1 to stable

## References

- Proto: `src/proto/dapr/proto/runtime/v1/dapr.proto` lines 71-76
- Dapr runtime promotion: https://github.com/dapr/dapr/issues/6582

markjgardner:
# Description

Dapr promoted the bulk publish API out of alpha in runtime 1.17, but the JS SDK still targeted the alpha1 variants on both transports. This moves `client.pubsub.publishBulk(...)` onto the stable API without dropping support for the runtimes people are actually on.

**Approach: stable first, with a transparent fallback.**

- gRPC calls `BulkPublishEvent`, retrying `BulkPublishEventAlpha1` when the sidecar does not serve it.
- HTTP calls `v1.0/publish/bulk/{pubsubName}/{topic}`, retrying `v1.0-alpha1` on a 404.
- The vendored proto backports the stable `BulkPublishEvent` RPC and marks alpha1 `deprecated`, matching upstream dapr/dapr, and the affected stubs are regenerated.

The public surface does not change. `publishBulk` keeps the same name, parameters and return type, so no consumer code needs updating. The downgrade is remembered per client, so a pre-1.17 sidecar costs at most one extra round trip rather than one per call.

## Detecting an old sidecar is not what you would expect

This is the one genuinely non-obvious part of the change, so it is worth calling out.

A sidecar that predates the RPC *should* answer `UNIMPLEMENTED`. It does not. Dapr installs a catch-all gRPC handler that forwards unrecognised methods to service invocation, so a pre-1.17 sidecar instead returns `UNKNOWN` with:

```
failed to proxy request: required metadata dapr-callee-app-id or dapr-app-id not found
```

An `UNIMPLEMENTED`-only check therefore never matches, and bulk publish over gRPC breaks on every runtime older than 1.17. This was caught by running the e2e suite against 1.16.12; unit tests with a mocked client cannot surface it.

Both shapes are recognised. The `UNKNOWN` branch is deliberately narrowed to the proxy-failure signature rather than treating every `UNKNOWN` as a missing RPC, because a proxy failure proves the sidecar never reached the publish path and a retry is safe, whereas a blanket match could retry a batch the sidecar had already accepted and publish it twice.

The HTTP side has a similar ambiguity: 404 is also how the sidecar reports an unknown pub/sub component. So the HTTP downgrade is only remembered once the alpha1 endpoint has actually answered, which is the only real evidence the runtime predates 1.17.

Heads-up for maintainers: this is sidecar behaviour rather than client behaviour, so the equivalent `UNIMPLEMENTED` checks in the Go, Python and Java SDKs look like they have the same latent gap.

## Why a fallback rather than bumping the pinned runtime

A hard switch is not really available here. The stable RPC is absent from the pinned v1.16.0 proto, so it would require bumping the pin, and Dapr 1.17 split `dapr.proto` into 14 files. That relocates message types out of `dapr_pb` into new modules and breaks imports across the whole SDK, which is a much larger change than this issue. Separately, e2e defaults to `DAPR_RUNTIME_VER=1.16.12`, so a hard switch would fail CI, and 1.16.x is still actively patched. Since 1.17 and 1.18 both still register alpha1 on both transports, the fallback never becomes dead code that silently breaks.

## Please look closely at the vendored proto

`dapr/proto/runtime/v1/dapr.proto` is hand-edited here. The sanctioned path is bumping `DAPR_VERSION` in `scripts/fetch-proto.js`, which is what #531 did for workflow, and I did not take it because of the 14-file split described above. The backported RPC is byte-for-byte identical to dapr/dapr, including `option deprecated = true` and the surrounding comments, so a later fetch at v1.17+ converges rather than conflicts. If you would rather see the pin bump land first as its own PR, that is a reasonable ask and I am happy to rework this on top of it.

## A note on the generated stubs

Flagging this in case it is useful rather than as a complaint, since it shaped how I generated the delta.

The committed stubs under `src/proto` were last generated on 2026-04-16 by the buf migration in #752, using protoc-gen-es v2.11.0 in ESM form. #826 then switched `buf.gen.yaml` to `js_import_style=legacy_commonjs` on 2026-05-26 without regenerating them. The result is that running `npx buf generate` on current main rewrites every stub from ESM to CommonJS, around 750 lines. That is your intended direction rather than churn, but it does mean any routine regeneration during an unrelated PR carries the migration with it.

So I generated the bulk publish delta in the existing ESM style, which keeps the codegen diff here to 43 lines across four files and keeps this PR about pubsub. Note the proto change is the load-bearing part either way: CI runs `buf generate` during prebuild, so the stable RPC is picked up from the `.proto` regardless of the committed stub format, which I verified with a local `npm run build`.

Happy to regenerate as CommonJS in this PR instead, or to send the catch-up as its own PR, whichever you prefer.

## Verification

E2E run through testcontainers, the same harness `test-e2e-testcontainers.yml` uses:

| Runtime | Result |
| --- | --- |
| 1.16.12 (pre-stable, exercises the fallback; current CI default) | 44 / 44 |
| 1.17.14 (serves the stable RPC, no fallback) | 44 / 44 |

Unit tests: 335 passing across 32 suites, including 13 new ones, 7 for gRPC and 6 for HTTP. Beyond the happy paths they cover the proxy-failure shape falling back, an unrelated `UNKNOWN` not falling back, a 404 from both HTTP endpoints being treated as an unknown component rather than an old runtime, and the downgrade being probed only once per client.

Total diff is 11 files, +459 / -16.

## Issue reference

Please reference the issue this PR will close: #786

## Checklist

- [x] Code compiles correctly
- [x] Created/updated tests


markjgardner:
@WhitWaldo Hope you don't mind me stealing your issue but #786 looked like an easy starter issue so I gave it a stab. Let me know what you think.

